PR #18, opened 2022-05-26
Add files via upload

jgcape:


breemclennan (commented):
Great work Julian and Ryan!
I really like the maps you have produced on green wall location suggestions.
Kind regards,
Bree

alex-anhvuong:
The map layers look great team! Also I like the last map where it's more articulated for non-technical use.
It's interesting how this one notebook is 17MB (maybe from the canopy data?).
Note for future: if web team can't convert this, try reduce map size by aggregating choropleth data, or lower dpi of map.

PR #25, opened 2022-07-26
Bicycle Route Network and Road Safety Part 2

breemclennan:
Bicycle Route Network and Road Safety Part 2. This is the second notebook from the use case on exploring bicycle route networks and road safety/accidents


hannahsm1th (changes requested):
Line 746 there is a typo "represending"

hannahsm1th:
Colours for the 'Count of Accidents by Atmospheric and Light Conditions' could be hard to distinguish from one another

PR #55, opened 2022-10-05
Abosede a

abosedealli:
Adding this Smart bin and disposal optimization,


breemclennan (commented):
Hi Bose,
Well done on creating this python notebook for the smart bin use case. You've covered some very intriguing content with the smart bin data.
I would suggest updating the file name for the python notebook to conform to the naming standards applied in the /usecases/ folder in the master branch of the repository. For example 'usecase-SmartBinDisposalOptimisation.ipynb'
Are there any future opportunities for further work which could be applied to this use case? It might be good to capture these in the notebook document or on the Trello card.

Kind regards,
Bree

hannahsm1th (changes requested):
Hi Bose, could you also move your documentation to the MOPCode/datascience/documentation folder, and update that filename too? The current name is something about blockchain. 
Thanks, 
Hannah.

abosedealli:
Ops,  thats terrible..am sorry.. will do that now

________________________________

hannahsm1th (approved):

PR #65, opened 2022-11-28
Update use-case.css

hannahsm1th:
Added in styling for custom classes in new template. See below for a screenshot.

The space between the sections is gone as I switched to using div tags instead. I think it looks better this way but let me know if you disagree, can always stick a margin in to get the old styling.

@m1ckw The darker colour is the one from the Figma that most closely matched your mock up, feel free to tweak the lighter colour if you think it needs it.

Once this is approved I'll update the existing published use cases with the new classes, and let the web dev team know when they are ready for reupload.

usecae-newcss

### DIFF
--- a/webapp/flaskr/static/styles/use-case.css
+++ b/webapp/flaskr/static/styles/use-case.css
@@ -22,6 +22,56 @@ code.hljs {
   Colors: /blob/v0.64.0/lib/css/exports/_stacks-constants-colors.less
 */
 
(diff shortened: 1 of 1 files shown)


m1ckw:
Looks great @hannahsm1th. So much nicer than black and grey.  I don't think any adjustment to the lighter colour is required, they look good together imo. Also, love the fact that you were able to remove the gap between the sections. Nice work!

calebwebb (approved):
Looks good to me.

PR #74, opened 2022-12-08
Update to Use Case css & NB Template

m1ckw:
Hi @calebwebb, a few small changes on top of @hannahsm1th recent updates. Added padding to the colour-filled header components and made an adjustment to the font sizing. Changes to both the css file and the use case notebook template.


calebwebb (approved):
Ok to merge.

PR #76, opened 2022-12-12
Jack analysis

jackpham206:
Uploading final version of High rise use case


hannahsm1th (changes requested):
This is excellent! The visuals are very informative. I have a few small styling requests:

Required:
- Please change the old headings to match the new template so the styling can be done by use case CSS
- Cell 51 has an error
Suggestion:
- Colorblind friendly bar/scatter plots (can use seaborn palette)
- Cell 26 - is it possible to change the scaling of the colour mapping? There are a lot of pale yellow blocks that have quite different values

jackpham206:
I have done all the requests above @hannahsm1th. I have also fixed the notebook style too.

hannahsm1th (approved):

PR #93, opened 2023-04-04
NewAPI

AdamBully10:
This request changes the API of an existing use case and one for publishing. It also updated the corresponding HTML files.


Bimsickle (approved):
API changed to V2.

Bimsickle:
API repointed.

PR #100, opened 2023-04-26
Steven - Complete repointing of new business location api

Stevenng7:
Hi Angie,

I've completed the repointing the api for the new business location. I've changed the 3 datasets from the old socrata api request to the v2 api using the url generated from the export dataset. This url is then turned into a pandas dataframe through pd.read_csv(url).


Bimsickle:
files outside of working folder have been edited

PR #113, opened 2023-05-15
Content Security Policy Changes

lkocon69:
As is, this won't disrupt the project. It is still report-only (which WILL need to change in time) but the CSP whitelist is more extensive in where it trusts its resources.

Area for improvement is using 'strict-dynamic' within the CSP and the HTML pages for all inline and eval events (which are present but commented out in the whitelist). If report-only is deactivated, all inline and eval events will be blocked unless they're accounted for in 'strict-dynamic', but that accounts for most of the warnings already in the project.

### DIFF
--- a/webapp/flaskr/__init__.py
+++ b/webapp/flaskr/__init__.py
@@ -66,7 +66,8 @@ def create_app(test_config=None):
              ]
              },
              content_security_policy_report_only = True,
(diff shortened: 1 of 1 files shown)


Bimsickle:
Rejecting as discussed with Lucas

PR #118, opened 2023-05-16
UpdateAdamB

AdamBully10:
Adds Tri 1 2023 use case - Disability Analysis along with HTML and JSON file

Further moves existing Tri 3 22 use case - Liveability Analysis along with HTML and JSON to ready to be published folder


Bimsickle (approved):
Looks good

PR #131, opened 2023-05-23
Published two usecase reworks and moved files in the datascience folders

TateRemzi:
- Published the rework for "New Business Locations" and "Evaluate Business Location..." 
- Moved relevant filed from REWORK COMPLETED into PUBLISHED for the aforementioned use cases
- Moved "Liveability analysis" and "Walkability" use case files from READY TO PUBLISH into PUBLISHED


Bimsickle (approved):
all good

PR #147, opened 2023-08-11
Update use cases

yashikaarora046:
Updated 3 use cases making it similar to new business location page


BarryCILs (approved):
Thanks Yashika

PR #267, opened 2023-11-30
Alison pull requests

AliColli:
Adjusted new version of laneways use case, finished API v2.1 code, finished laneways planting user interactive case.


SijuDENG (changes requested):
Your .gitgnore have your local path. Also does not conform to the old format.

SijuDENG:
you have a change on  .gitignore  file under MOP-Code folder.  Please make sure not to include your local path, such asC: or 764/MOP-code or the like. Please also comply with the old format. Or please cancel the changes. pull as a whole cannot selectively agree. I can't agree to the pull until you fix the problem.

PR #268, opened 2023-11-30
Update usecase_Bus-stops_Pedestrains_TEMPLATE.ipynb

Shilpa337:
I updated the use-case template with a bit of introduction of my use-case.


SijuDENG (approved):
Thank you

PR #273, opened 2023-12-03
Alison

AliColli:
Use cases final draft, API converted to user entry. git-ignore local deleted comments. API v2.1 updates to API user entry, git frequently used commands finished, PDFs of documents.


AliColli (commented):
Deleted comments in git ignore and changed API to not have folder stored (updated personal API code as well so no data leakage)

PR #277, opened 2023-12-05
Add the file into use case 

jenniferyau:
Hi Team, please kindly review see if it is okay to merge, Thanks.


SijuDENG (changes requested):
Location: We are in usecases/ready to publish/T3 2023/
Folder: There should be a folder with the same name as the case
We use .json not . py
Case: good

jenniferyau:
I will delete this pull and recreate a new one. thanks

PR #278, opened 2023-12-05
Alison

AliColli:
Finished use case and placing in usecases>ready to publish, finished documentation and placing in data science>documentation folder (onboarding, use case publishing update and git commands cheat sheet)


SijuDENG (changes requested):
Case Location: We are in usecases/ready to publish/T3 2023/
Case Folder: There should be a folder with the same name as the case
.json is a iintroduction not output as html. Reference demo

AliColli:
Updated .json

PR #306, opened 2023-12-09
City living comfort 

SijuDENG:
Rewrite case Walkability map of the City of Melbourne became City_Living_Comfort 
add 3 files in ready to publish


RuofengQIU (approved):
thanks SIJU

PR #368, opened 2024-03-26
Use Case to be Retired

molliefernandez:
This Use case uses external data not part of the CoM datasets which is not publicly available and we seem to have lost the data anyway. Use Case should be retired.


molliefernandez:
[Pedestrian Safety Part 1 - RETIRED](https://trello.com/c/BnrtrtcU/196-pedestrian-safety-part-1-retired)

leakydishes (approved):
As discussed in Meeting Week 4 - approved

PR #381, opened 2024-04-01
Molliefernandez dependencies

molliefernandez:
In this pull request:
- Moved all dependencies to DEPNDENCIES folder
- There may be more to move, will keep looking
- Removed any unnecessary files/images
Next step is to update all dependency paths - will combine this with updating API paths for use cases


leakydishes (approved):
Nice work -- agreed pngs are no use.

leakydishes:
Approved, you can merge @molliefernandez

PR #400, opened 2024-04-08
Samiha haque

Samiha-Haque:
Hi @leakydishes  and @AliColli . I updated the legend in the graphs and completed the use case.


AliColli (approved):
Looks all good :)

PR #402, opened 2024-04-09
(WIP) Customer Service Analysis

carissawong:
The code includes fetching the data, data cleansing, aggregation and some visualization.

### DIFF
--- a/Playground/WingsumWong_Carissa/WIP_CS_Analysis.ipynb
+++ b/Playground/WingsumWong_Carissa/WIP_CS_Analysis.ipynb
@@ -3,7 +3,10 @@
   "nbformat_minor": 0,
   "metadata": {
     "colab": {
-      "provenance": []
+      "provenance": [],
+      "collapsed_sections": [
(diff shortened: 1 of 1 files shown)


HarleyNgo (commented):
I wonder if you want to run again the notebook so that the visualizations appear, as for now it does not
I'll approve if you just want to keep it as it is now.

carissawong:
Hi Harley, I already re-run the code and the visualizations should be appeared now.

AliColli (commented):
Hi Carissa, plots are still not rendering in GitHub, maybe see if there is another option you can do? Usually, they just leave a blank space where they should render in if they won't load. Also consider using .head(3) when checking dataframes or a shape function? Just to reduce the length of the use case in outputs.

HarleyNgo:
it still does not show visualizations but i still approve for this time
let's find a way to solve it next pull request

HarleyNgo (approved):

carissawong:
Thank you Harley! I'll figure it out in the next pull request!

PR #403, opened 2024-04-09
Added Jest test for Footer component

thamashag:
This initial commit introduces a Jest test for the Footer component. The test verifies that the Footer component renders correctly with the expected content. Additionally, Jest configuration was updated to resolve errors encountered during testing, including setting the test environment to jsdom and installing necessary dependencies. As a result, all tests, including the newly added Footer test, are passing successfully.


kaseyuu:
Please create a new pr after you fix the babel compiler issue :)

PR #407, opened 2024-04-09
Sprint 1 - Banner Poles Booking Use case

chihoang239:


AliColli (approved):
Great start on use case. When printing the data frames consider using .head(3) to only print a few rows (a future change, its fine for now :) )

PR #412, opened 2024-04-10
Raman

Raman1509:
This commit contains use case proposal, data fetched and cleaned, analysis needs to be tuned and part of the code needs debugging, plotting done but can be improved much more


AliColli (approved):
Great start on use case 👍

PR #419, opened 2024-04-11
Build-Render two user cases

kaseyuu:
Set API and constructed the file path. Parsing two old user cases to case study 1 and case study 2 for example. Will update this part after the database is ready.


Sakif-Hasan (approved):

kaseyuu:
[Database render](https://trello.com/c/ZUnfJC8v/499-database-render)

PR #432, opened 2024-04-14
Te claire - updated use case eco-metrix (in progress)

leakydishes:
Updated PCA code, monthly median value probe measures unser interface, time series, Pearson correlation and spearmans rank correlation, Granger Casuality test and Integrated data board using company colours.


AliColli (approved):

molliefernandez (approved):
You do too much work Te' Claire! Great Job!

PR #438, opened 2024-04-16
API NEW -Litter Traps & BBQ & Differing Site Typologies

TharushaCao:
Updated any errors faced by api new , all code works now with new API

- Updated use case Litter Traps & BBQ & Differing Site Typologies with New API Method
- Checked if the code still works 
- Fixed to make code work


TharushaCao:
[New API - Litter Traps & Public BBQ](https://trello.com/c/UnQO9sKS/520-new-api-litter-traps-public-bbq)

AliColli (commented):
As this is a finished use case maybe an explanation of the linear proof at the end would be good? Just explaining what it finds and why its needed? Also any references for this? (if not ignore that comment :) )The rest looks great!

twar9432 (approved):
API looks good. Just mirroring what Allison said, maybe some additional comments (I realise that was our use case so sorry/thanks)

AliColli (approved):
Approved and can now merge :)

PR #442, opened 2024-04-20
WD| Patch| fix page routing

kaseyuu:
Changing the page routing from case studies to user cases to be consistent with new header, so that case studies page can show again.


Tyrepo1 (approved):
Looks good

PR #444, opened 2024-04-21
WD_Use_cases_Adrian_Weerasinghe #PR2 [FIX]

adrianweerasinghe97:
Made changes for the database, page, searchbar, pre…view and usecase files. Updated the css file
updated the front end to use Next.js native routing instead of "react-router-dom"


Tyrepo1:
Please fix the background

PR #452, opened 2024-04-23
Create Job-Forecasting

s220313209:
The code file for use case Job Forecasting by Bao Ngo


HarleyNgo:
Hi Bao.
Please add description for the usecase.
Also, your file is not a ipynb one right? please make sure it's a ipynb so that we can see the output.

yashdawaree (approved):
Add Use Case Name and Description

PR #469, opened 2024-04-26
StatisticPage_UI_DEV_VictorXiao #PR1

Azureshadow06:
Statistics page UI -development] https://trello.com/c/2OEiIsAi/541-statistics-page-ui-development-max-2

Created three filter drop lists on statistic page


Tyrepo1 (changes requested):
image
Can you make it so that there's no blackbox at the bottom of the page

PR #470, opened 2024-04-26
Completed use case with tensorflow model

leakydishes:
c. Linear Regression Model makes a set of independent predictions on consecutive time steps. 
Fast Fourier Transformer (FFT) Test confirmed temperature was a predictor (normally used for sound/ wave lengths but can be used for temperature variance metrics).
Violin plots to see distributions of normalisations.
Predictions based on a window of consecutive samples from the data
Used Tensorflow model tf.keras model which has built in functions for deep learning models and useful for making time predicitons. 
Created a base model to predict the next value in the time series data (input window)
Created a linear model and examined ‘single step and wide/ spread window predictions’. 
Tested Dense, multi-step and convolution layer models, however linear regression with weights based on temperature outperformed other models with closest stats to base model). 

Linear model showed high accuracy and consistency across predictions y (0.63 ~ base model) with non significant diviations of 0.01 predictions.


leakydishes:
Need to fix issue in markdown

PR #472, opened 2024-04-27
finalising use case

carissawong:
updated use case (background, user story, conclusion), added more visualisation, include static plots for each of the interactive plots to see if this can be displayed properly on GitHub


leakydishes:
Hi @carissawong you need to fetch main as you have a deployment error and reviewers will not be able to merge. 

> 
Screenshot 2024-04-27 at 7 12 22 pm

PR #481, opened 2024-04-28
Simranjitsingh

sandhusimranjitsingh:


kaseyuu:
Hi Simranjit, thank for fixing the images. The only issue is that when there are quotation marks such as '' in your text, the system may not recognize whether they are part of the text or part of the program. The solution is to replace ' with &apos;.

kaseyuu (changes requested):
Hi Simranjit, thank for fixing the images. The only issue is that when there are quotation marks such as '' in your text, the system may not recognize whether they are part of the text or part of the program. The solution is to replace ' with &apos;.

PR #490, opened 2024-05-01
DS - Block Energy Consumption Use Case

twar9432:
Created new use case. Code includes:
 
 * Code and mark-ups in usecase template
 * Data import, preprocessing, analysis, and modelling


TharushaCao (approved):
Looks good @twar9432 

recommend adding an actual test of the neural model, maybe a user input for model prediction.

PR #493, opened 2024-05-02
StatisticPage_UI_DEV_VictorXiao_#PR3

Azureshadow06:
Statistics page UI -development] https://trello.com/c/2OEiIsAi/541-statistics-page-ui-development-max-2
- Created Barplot for Popularity Range
- Created filter drop list for years
- Created filter drop list for trimesters
- Fixed the issue that search function, now allowing users to search regardless of the case used.


abhishek-chiluka99 (approved):
The pull request seems to be good,

- there are no any code merge conflicts.
- I have checked the preview and its working good as the design.
- Checked all the functionalities working fine for the filter and table functions
- Just have a second review regarding node modules changes.

Everything is fine

kaseyuu (changes requested):
Please do not check in node_modules because that will increase hundreds of unnecessary changes. Node_modules doesn't need to be added cause it's already in gitignore. You can pull from the main before applying your changes and create a new pull request. Thanks.

PR #495, opened 2024-05-02
Data Science Melbourne's Biodiversity usecase 100%.

francisrusli:
Completed Melbourne's Biodiversity usecase.


leakydishes (changes requested):
Great to see you add 'Benefits and Conclusion' in mark down. This makes it very clear to your audience. 

Re: Filling missing values for categorical data with 'Unknown'. 
It is a good idea to calculate percentage of unknown values and then make a decision on how to approach the fill. 
## Future use cases: 
- columns with missing small records (217 or 34) vs larger (403). You can deal with them differently. Smaller missing values could potentially be removed from dataset and others could be approached with Forward Fill or using LLM to find the 'genus' based on phylum, class ect. 
- At the end of the day it is what ever is applicable to the use case. 'common_name' appears to be 'unknown' for many taxa, I would remove this. 

## Histogram:
Consider buckets of different Phylum, (unknown values) is 5th largest out of 34 class records. This distorts the dataset. 

## Taxa Distribution 
Your 'Taxa' has overlapping titles. Can you change code to:
'''
# Plot a histogram for the distribution of different taxa
...
plt.xticks(rotation=90) 
...
plt.show()
'''

## Inspect dataset for taxa
- Folium map is not showing
Make this Notebook Trusted to load map: File -> Trust Notebook

## Mapping of bats based on their unique locations
- Folium map is not showing
Make this Notebook Trusted to load map: File -> Trust Notebook

## Heatmap
- Folium map is not showing
Make this Notebook Trusted to load map: File -> Trust Notebook

## Clustering the Bat Data
- Folium map is not showing
Make this Notebook Trusted to load map: File -> Trust Notebook

## Family histogram
- Consider creating buckets based on a second lable as the bar chart is difficult to read

Great Temperature, Humidity distribution! 

## Temperature and humidity boxplot
- Graph titles are overlapping and I cant read them. Consider buckets to categorise data

## Mapping all data points
- Folium map is not showing
Make this Notebook Trusted to load map: File -> Trust Notebook

## Creating Clusters
- Folium map is not showing
Make this Notebook Trusted to load map: File -> Trust Notebook

## Heatmap
- Folium map is not showing
Make this Notebook Trusted to load map: File -> Trust Notebook

@francisrusli please review above changes, you can disregard changing null values as this a suggestion. We need to clearly be able to read maps.

francisrusli:
@leakydishes Hi Te!
I have fixed all but one of the problem. regarding to the Folium maps not showing, When running in my local vscode the map is visible however it is not in github. 

the only solution i can think of is saving the map and uploading it to the usecase folder. Is this a proper solution or should i just leave it as is?

Thanks!
Francis

PR #504, opened 2024-05-03
Added Visualisation

Raman1509:
This commit contains visualisation added to the usecase.


molliefernandez (approved):
Hi @Raman1509 
Your use case looks great, I love the parts at the start where you have explained what you are doing, and have continued this throughout the use case. 
You might want to remove the printing of the multiple locations, it takes up most of your use case and probably isn't necessary. 
Overall great job.

PR #505, opened 2024-05-03
Pedestrian and bus stops 100% completed

remmzytom:
all process involved

- Data manipulation and analysis using pandas
- Advanced visualization techniques with matplotlib and geopandas for geospatial data analysis.
- Spatial analysis using geopandas to understand geographic relationships between pedestrian activity and bus stop locations.
- Implementation and interpretation of clustering algorithms for identifying spatial patterns in data.
- Integration of machine learning techniques (such as clustering) with spatial analysis to derive actionable insights for urban planning.
- implementation of GAM and polynomial regression and LASSO regularization for non-linear relationship
- Application of data-driven decision-making processes in urban planning contexts.
- Good recommendation for urban planning


TharushaCao (changes requested):
- remove the brackets in the use case description section

-{calling the three datasets-----
pedestrian counts dataset-----
bus stop datasets-----
sensor reading datasets----} Remove this if not applicable 

- Please use the **Preferred Method**: Export Endpoint only once as a function and call it multiple times to get datasets, lets try not to repeat code .

- Conclusion is needed at the end if usecase is finished

molliefernandez (approved):
Hi James, 
Great work on your use case! The only thing I would say is that some of your text between the code is set to heading style when it should be body text. Also please make sure you include a conclusion if your use case is completed.
Otherwise great job!

TharushaCao (approved):

PR #508, opened 2024-05-03
Real Estate (data science) Sprint 2 pull request

Pnishadini:
This use case is 80% completed. Thank you!


molliefernandez (changes requested):
Hi @Pnishadini Prathiba, Great use case, a couple of things to fix when completing it:
- Your Scenario should not be in first person exactly. It should be in the perspective of someone who is an urban planner. For example instead of saying "I am going to analyse....." Start with: "As an urban planner I am going to analyse.... because..... to make decisions about......."
- I would love to see more explanation between your code about what you are doing and why etc. this is meant to eb a tutorial for people to learn how to use the datasets. 
- Some minor spelling and grammar errors
- The results section looks great!
- Don't forget to add a conclusion when you are done and any references :)

Other wise looks great, can't wait to see the finished product :)

PR #513, opened 2024-05-03
New version of use case Job Forecasting 

s220313209:
The new edition of use case Job Forecasting in Melbourne, including Random forest model and map using Folium.


leakydishes (changes requested):
Screenshot 2024-05-04 at 12 42 55 am

You currently have more then 8+ branches.
Please compile or delete as we can't approve merges until you do.

Your use case is empty?
Screenshot 2024-05-04 at 12 43 56 am

@s220313209

PR #514, opened 2024-05-03
WD_BUILD_search content option

kaseyuu:
Added a filter to choose from search by title or search by content. Built search by content function for user cases page.


Sakif-Hasan (approved):

kaseyuu:
[Search content function - User Case page](https://trello.com/c/0TVLVI4e/565-search-content-function-user-case-page)

PR #515, opened 2024-05-04
DATA SCIENCE TEAM - Tree History - 100% Complete

hiruboy96:
**_Use case completed_**

Use case includes:

Adding Description
Adding Title
Loaded Data through API
Removing Unwanted columns
Renaming columns to match the other dataset
Historical vs. Current Species Diversity (Venn Diagram)
Change in Species Abundance Over Time
Geographical Distribution of Trees (Map Visualization)
Time-Series Analysis of Tree Planting
Habitat Analysis for Pre-Colonial Species
Ecological Dynamics of Tree Species in Parks vs. Streets in Melbourne
Adding filters to the map
Timeseries Model Fitting (ARIMA Model Forecasting)
Forecasted tree plantings for the future 5 years
Updated the API to catch the data by hiding the API key using Environment Variable.
Finalizing the use case by achieving the desired outcome.


HarleyNgo (approved):
Great work!
A few things I need the use case can be improved:
- Code explanation is too detailed :) -> I suggest just explain the output of the code block/why it helps to the overall use case.
- Only print top "10 or 20" for Ecological Dynamics of Tree Species in Parks vs. Streets in Melbourne as it's too long and unuseful
- use library command "ignore warning" to remove warnings (better visuals and flow)
- about the mode --> comment on the model's result and how/what/why to improve if possible

hiruboy96:
Will be uploading new modified files

PR #523, opened 2024-05-04
DATA SCIENCE TEAM - MELBOURNE JOB FORECASTING NEW USECASE - 100%

s220313209:
This use case include:
+ Use case introduction
+ Import data using API
+ Data analysis for 4 datasets
+ Choosing the most suitable dataset to apply model
+ Apply Random Forest to forecast the number of job for different areas in Melbourne.
+ Using folium to visualize the map.


HarleyNgo (changes requested):
A few things to help your use case have better flow:
- import all your libraries on the top
- only print df.head() to reduce the length
- no need to always print the df after you do any pre-processing
- your result of MSE of Random Forest is high, do you consider optimize it?
- have headers so that the users can follow easier
- add explanation why you have those visualizations and how it helps in your use case

s220313209:
> A few things to help your use case have better flow:
> 
> * import all your libraries on the top
> * only print df.head() to reduce the length
> * no need to always print the df after you do any pre-processing
> * your result of MSE of Random Forest is high, do you consider optimize it?
> * have headers so that the users can follow easier
> * add explanation why you have those visualizations and how it helps in your use case

Sure I will update the code as your suggestion.
For those with no head, I just want to see more data then I can get more insight from them.
The data seems not good for applying model, I will modify a little bit.

PR #527, opened 2024-05-05
DS API repoint Liveability analysis Dependency 

Samiha-Haque:
Hi @molliefernandez and @TharushaCao I am trying to do API repointing of Liveability Analysis usecase which had a dependency dataset. I moved it to the DEPENDENCIES folder.


molliefernandez:
Looks Good @Samiha-Haque, maybe change the name of the file to include the use case it is associated with. Or even put it in a folder with the use cases name as well as changing the file name so we know exactly what it is for.

molliefernandez (changes requested):
Looks Good @Samiha-Haque, maybe change the name of the file to include the use case it is associated with. Or even put it in a folder with the use cases name as well as changing the file name so we know exactly what it is for.

TharushaCao:
Looks good , apart from changes from mollie all good

TharushaCao (approved):

Samiha-Haque:
hi @molliefernandez and @TharushaCao, I hope the above commits are good to be merged?

PR #538, opened 2024-05-10
StatisticPage_UI_DEV_VictorXiao_#PR3

Azureshadow06:
[Statistics page UI-development] https://trello.com/c/2OEiIsAi/541-statistics-page-ui-development-max-2

- Created Barplot for Popularity Range.
- Created filter drop list for years.
- Created filter drop list for trimesters.
- Fixed the issue that search function, now allowing users to search regardless of the case used. 
- Fixed the issue that the plots are n not centered.
- Fixed the issue that the search bar doesn't have full width of below table. 
- Extended the distance between body and footer.

### DIFF
--- a/next_webapp/src/app/statistics/page.tsx
+++ b/next_webapp/src/app/statistics/page.tsx
@@ -29,15 +29,15 @@ import {
 const Statistics = () => {
     // Dummy Array
     const caseStudies = [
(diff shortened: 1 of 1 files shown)


kaseyuu (approved):
Looks good！

PR #542, opened 2024-05-11
DS Public Event Planning usecase 100% complete

Samiha-Haque:
Hi @AliColli and @RhuthK 
I have completed(100%) the Public Event Planning usecase and moved it to the READY TO PUBLISH folder with 3 files (.ipynb, .json, .html)
Let me know if any changes are needed.


Samiha-Haque:
[DS New Use Case - Public Event Planning](https://trello.com/c/nJNTl2MA/489-ds-new-use-case-public-event-planning)

AliColli (approved):
3 files look good to me :)

PR #567, opened 2024-05-13
WD_Frontend_PL

plhagyal:
Upload Page
-styling fix
-responsive css


next_webapp/public/styles/CasestudiesBackup.css
adrianweerasinghe97, line 1: The container class has been defined  twice, once commented out and once active. 

The initial comment starts with /* but does not have the proper closing */. This could cause the rest of the styles to be treated as comments, rendering them inactive.

adrianweerasinghe97 (commented):
In the casestudiesbackup.css file
The container class has been defined twice, once commented out and once active.

The initial comment starts with /* but does not have the proper closing */. This could cause the rest of the styles to be treated as comments, rendering them inactive.

kaseyuu (changes requested):
Hi Pema, when there are quotation marks such as '' in your text, the system cannot recognize whether they are part of the text or part of the program. The solution is to replace ' with &apos;.

plhagyal:
> Hi Pema, when there are quotation marks such as '' in your text, the system cannot recognize whether they are part of the text or part of the program. The solution is to replace ' with '.

I haved made the changes. Thanks for the feedback

plhagyal:
deployment issue has been fixed

kaseyuu (approved):
Looks nice :)

PR #568, opened 2024-05-13
Usecase: House Prices, Landmarks and Places of Interests

yashdawaree:
Updated with API and Explanations.


AliColli (dismissed):
Yash make sure you move your file into the update needed folder if it is not complete or into the ready to publish if it is completed :)

Otherwise looks good to me!

AliColli (changes requested):
Hi Yash! Make sure you move your use case into the correct folder. Than I will approve as it looks good :)

PR #569, opened 2024-05-13
DS team. Pedestrian Prediction. 100% Complete, sprint 3 pull request 

trostov:
All 3 files included (ipynb, html, json) 
Use case includes:

- Updated use case introduction
- Improved Dataset cleaning/merging
- Widget based visualisation
- Static median visualisation for hour/day/month pedestrian values.
- Creating a numerical dataframe for ML use
- Regression models trained on and predicting the values of a single sensor in the dataset, with performance results
- Training a list of models, one for each sensor, involving the whole dataset
- Training a composite model using encoding
- Folium visualisation for both of the above


AliColli (approved):
Use case looks great and all 3 files in ready to publish. Ready to merge from me :)

HarleyNgo (approved):

PR #578, opened 2024-05-13
DATA SCIENCE Real Estate

Pnishadini:
Sprint 3 pull request


molliefernandez (approved):
Looks good. 
Only thing I would suggest for change is for you to change to Australian English instead of American English. E.g. "analyze" should be "analyse". This is for an Australian council office, so it should be in Australian English.

Pnishadini:
will fix it

PR #581, opened 2024-05-14
DS USECASE : HOUSE PRICES, LANDMARKS AND PLACES OF INTERESTS UPDATE NEEDED

yashdawaree:


leakydishes (changes requested):
@yashdawaree - 
What a great use case, I like the way you have both created 'Results obtained by fitting a regression model into the housing data' 

Please fix:
File needs to be in a directory.
1. Create a directory (folder) 'house_prices_landmarks_places_interest'
2. Upload your .ipynb file into this directory
3. Re-request

Thanks!

PR #596, opened 2024-05-19
Files added to folder

Pnishadini:
@twar9432 files have been added to the folder


twar9432 (approved):
@Pnishadini Approved, however I would consider giving it a more descriptive name first. "Real Estate" is very general.

PR #608, opened 2024-07-26
Create a new folder in the playground

samj001:


samj001:
@Samiha-Haque @HarshDW09 
Hey guys, would it be possible for you to help review my pull requests? Thanks!

Samiha-Haque (changes requested):
Hey @samj001 you need to use API 2.1V to load datasets. Remove all folders from playground and move your use case folder to T2 2024 in READY TO PUBLISH. You need to only keep the ipynb file, the html file and json file and delete all other files which aren't storing important dependencies.

PR #655, opened 2024-08-07
Data Science Team - Social Indicator Changes: 60% complete

trostov:
- Imported the various datasets
- Cleaned and formatted the data
- Created an interactive visualization
- Created manual visualizations regarding covid impact
- Averaged results and grouped by respondent group for second part
- Created initial interactive visualization to compare demographics


francisrusli (approved):
Looks good thomas :)

remmzytom (approved):
Your work looks great, but going forward, it would be even better if you could add some explanations about your visualizations. This will help others understand the insights you’re trying to convey

PR #660, opened 2024-08-07
Data Science team - Public Transport Accessibility to Business Locations such as Bars/Taverns/Pubs - 70% Complete

VaishaliThakoor:
Usecase for Sprint 1 includes:

- Use case proposal
- Importing modules and fetching data
- Exploratory data analysis with data cleaned and some visualisation of the data, further visualisation would be added
- First part of finding all the nearest transport service to the business locations

Future work:
Trying to provide recommendation based on the findings


trostov (approved):
Hi Vaishali,
Use case looks good, however the use case is 45MB, too large to display on github, so please see if its possible to reduce this size.
Also I would recommend adding titles to the graphs, and using the standard v2.1 api data collection, unless this isnt possible for your datasets.

ChathuzCodeCrafter (approved):
Hi Vaishali,
 Greate work, however, please check the possibility to reduce the size as it motioned as the file is too large to display in Github. I suggest you add the tittles to the graphs in your use case if possible.

VaishaliThakoor:
@ChathuzCodeCrafter @trostov thank you so much for the review! I will add the titles. The usecase is too big to display on github because of the dataset size which I have already reduced and that is why i didn't use the v2.1 API data collection as that would make the dataset even bigger and for the moment i think the best way to view it is using nbviewer until i figure another way around it.

PR #676, opened 2024-08-08
AI TEAM - Two Axle Trucks Data Annotation

balexiadis:


lkocon69 (approved):
After earlier reviews for 632 and 674 and those changes implemented, I approve this inclusion

PR #687, opened 2024-08-09
Merging Lucas' branch containinf Three axle Articulating category samples

lkocon69:


lkocon69:
Kevin has noted that the label index was incorrect, all label files have a value of 0 instead of the intended label of 15

PR #689, opened 2024-08-09
Three Axle Articulating

lkocon69:
Labels all edited to contain the label index of 15, per Kevin's review


Kevinbui16 (approved):
All good!!!

PR #713, opened 2024-08-11
Images and Annotations of 2 Axle Bus

Krieg-117:


lkocon69:
Personal message prior to recent commit, noted that the label files all had the incorrect index of 0 (sedan) instead of the intended index value (11) - this has been corrected

lkocon69 (approved):
All valid images of two-axle buses, and correct label files

PR #714, opened 2024-08-13
Train and validate  Health Behavior AI model

Sahana123456sa:
The model is about achieving better accuracy upon training for analysing health behaviour dataset using random forest approach and evaluated using gridsearch.


indigovn (approved):
• The code does not check or handle missing values. It's good practice to check for and handle them before processing the data.
• The age binning is done correctly, but you might consider making the age groups more flexible by dynamically calculating the bins or using a different binning strategy (e.g., quantiles).
o Quantile Binning: Best for handling skewed data and ensuring each bin has a roughly equal number of observations.
o Custom Binning: Ideal when you have specific age ranges that are meaningful for your analysis. For instance,
bins = [0, 18, 25, 35, 45, 55, 65, 100]
labels = ['0-17', '18-24', '25-34', '35-44', '45-54', '55-64', '65+']
data['age_group_custom'] = pd.cut(data['age_numeric'], bins=bins, labels=labels)
o K-Means Clustering: Useful when you want data-driven age groups.
o Equal Width Binning: Good for treating age as an ordinal variable with equal intervals.
• You are using LabelEncoder for multiple categorical columns. While this is okay for binary or ordinal categories, it might introduce unintended ordinal relationships in location. You could consider OneHotEncoder if these categories are nominal (unordered).
• Model evaluation: Consider adding other metrics for model evaluation, especially if you have imbalanced data, such as precision, recall, F1-score, or AUC-ROC.
• Cross validation: Although you perform cross-validation within GridSearchCV, you also manually perform cross-validation after selecting the best model. It’s typically not necessary to do this again unless you are performing additional validation or comparing models.
• The GridSearchCV automatically refits the best model on the entire dataset after finding the best parameters, so there’s no need to manually refit.

Let me know if you have any questions/clarification

Cheers,
Alex

PR #715, opened 2024-08-13
Daljeet

PDALJEET:
use case description-
Usecase includes in scenario
other thing is what you can learn from the usecase 
introduction
and i did cleaning and found the non null values .


PDALJEET:
use case description-
Usecase includes in scenario
other thing is what you can learn from the usecase
introduction
and i did cleaning and found the non null values .

PDALJEET:
having some issue need to fix it will create new one

PR #721, opened 2024-08-15
Mptbot

indigovn:


luilding (approved):
Code looks great, solid foundation for the chatbot files and offering functionality via the actions.
Approving but I do have some queries/points:
- Are you seeing any cases that the bot gets confused and executes the wrong action/doesn't execute the action at all? If so we should construct a readme explaining these
- I noticed the dataset_path is pointing to your local machine, I recommend using a relative path to ensure the code works in different environments.

PR #732, opened 2024-08-21
Full implementation for UC 1-4 and 13

indigovn:
This code will work for the following usecases
**Basic Route Planning**
"What is the best route from [Station A] to [Station B]?"
"How do I get from [Station A] to [Station B]?"
"Show me the fastest route from [Station A] to [Station B]."
**Schedule Information**
"What time does the next train to [Station B] leave from [Station A]?"
"When is the next train from [Station A] to [Station B]?"
"What are the train schedules for [Station A]?"
**Transfers and Connections**
"Do I need to change trains to get to [Station B] from [Station A]?"
"How many transfers are there between [Station A] and [Station B]?"
"Is there a direct train from [Station A] to [Station B]?"
**Route Optimisation**
"What is the quickest route to avoid delays?"
"Which route has the least number of stops from [Station A] to [Station B]?"
"Recommend a route to avoid Zone [X]."

Please have a look and merge to the master.

Thanks...
Alex

### DIFF
--- a/artificial-intelligence/chatbot/code/mpt_bot/actions/actions.py
+++ b/artificial-intelligence/chatbot/code/mpt_bot/actions/actions.py
@@ -3,59 +3,74 @@ import folium
 import os
 import pandas as pd
 import logging
-from typing import Any, Text, Dict, List
+from typing import Any, Text, Dict, List, Optional
 from fuzzywuzzy import process, fuzz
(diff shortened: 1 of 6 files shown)


luilding (approved):
Good work Alex. Approving the changes to RASA yaml files and to the new feature actions in actions.py

PR #735, opened 2024-08-23
MOP_WD_Adrian_Weerasinghe_PR2 [fix]

adrianweerasinghe97:
Nested div and h1 Tags:
The code had redundant div and h1 tags that are not properly closed or are nested unnecessarily.

For example, there was an extra h1 inside another h1, which is not valid HTML.

Multiple Unclosed Tags:
There were unclosed tags or tags that are closed incorrectly.

For example, there were h2 tags opened multiple times without closing the previous ones.

Repeated Sections:
The structure of the content is repeated. This repetition includes unnecessary duplicate divs and text, leading to cluttered and hard-to-maintain code.

Inconsistent Classes:
There were inconsistencies in the class names, which might cause styling issues.
For example, the class names like text-[40px] and text-[20px] are applied in multiple places, which might cause confusion about the intended font size.

### DIFF
--- a/next_webapp/package-lock.json
+++ b/next_webapp/package-lock.json
@@ -88,12 +88,12 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.24.2",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.24.2.tgz",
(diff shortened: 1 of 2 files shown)


Tyrepo1 (approved):
looks good

PR #736, opened 2024-08-24
images of sedan boat trailer and caravan.

qasimzia1:
adding images  of sedan boat trailer and caravan.


lkocon69 (changes requested):
There are multiple issues between multiple task submissions that need changes before approval.
- The boat category is not about classifying boats in the water, but about cars towing boats. This will need a complete redo
- There are 18 caravan images that are generated by DALL-E that have a long filename that is causing issues with Github distribution. These images and label files need their names shortened, and then reconsidered for the next change below.
- Just like the boat category, it's about the cars towing caravans and campervans instead of just the caravans. Samples with a car towing a caravan can be kept but the rest need to go.
- Sedans seem fine but made up of only photos taken personally. While I admire the effort, you have the rest of the internet to also collect other "cars" to use as images.
- Trailers consist of images containing only trailers, when again the purpose is to have images with cars towing trailers.

lkocon69:
Altogether pretty good, though some changes will be needed once integrated:
- samples to remove that are not applicable: caravan images348, images361, images409, images482, pic94, pic104, pic107, 
- samples to change category: caravan- images383, images384. caravan-weight-og, images12, images27, images87; boat- pic17, pic18, pic72, pic86, 
- images415 can also be tagged for trailing a boat
- while resized image duplicates are fine for training (where due to dimension compression they are different enough), straight-up duplicates of images will need to be removed to limit sample biases
- trailer samples remain problematic where most of the samples aren't attached to cars as implied by the project requirements (perhaps not by the task, but certainly by the project)

lkocon69 (approved):

PR #759, opened 2024-09-02
WD_Frontend_Chaya#PR6

chaya221071557:
Login Page Multi-language functionality.


Tyrepo1 (changes requested):
Add the header to the logion page and use the language button in the header instead of adding it to the login screen

chaya221071557:
okay. I will make the changes and submit a new Pull request.

PR #765, opened 2024-09-03
Fixing About-Us Responsiveness

koiCoding:
Fixing position and size of content within about us page to relative measurement units along with breakpoints for each screen size using Tailwind CSS.


Tyrepo1:
The design was changed, message Muhammad about this

PR #767, opened 2024-09-04
Ziqinzhao

ziqinzhao2024:
Hi,this is Ziqin Zhao. I upload all images and labels include Double Road Train , MAD , B double, Heavy Truck Tralier , Three Axle bus, Three Axle Truck. All together 1020 images and labels in total.


lkocon69 (approved):
Kept my eye on its progress before pull request. Content is good

PR #768, opened 2024-09-04
Alex1

indigovn:
**Major update:** 
1) Loading dataset directly from the live source. 

2) Integrated with the map generation. the map will be generated right after the trip information and use able to click on the link to see the route map details"


Samiha-Haque (approved):
@indigovn  Looks good! Approved.

PR #781, opened 2024-09-09
MOP_WD_Hoang_Duy_Vu_#PR5 

Tyrepo1:
- Redesigned Use Case page according to new design
- Added T1 2024 Use Cases to repo for display
- Remade the display and search api to display the new use cases


adrianweerasinghe97 (approved):
good

PR #782, opened 2024-09-09
Licensing Page:context update and background color change

s217034263:


Tyrepo1:
Various conflicts with other files and infinite files changed
image
image
Should probably make a new PR

PR #785, opened 2024-09-09
Delete Playground/Adersh/usecase_TEMPLATE.ipynb

AntonyJr321:
Unwanted File 

The file was just the rough model so had to delete it


francisrusli (approved):
@AntonyJr321 removing usecase template file. No issue.

PR #786, opened 2024-09-10
Web_Design_Abhishek_PR2

abhishek-chiluka99:
- Re designed the theme of dark theme of home page with suitable color scheme. 

- Designed Use Cases Mobile view page.

- Designed Use Cases File Mobile View page


Tyrepo1 (changes requested):
You made 2 pull requests with the same content. Please remove one of them

PR #787, opened 2024-09-10
Web_Design_Abhishek_PR2

abhishek-chiluka99:
Redesign the Home Dark theme page with suitable color scheme, 
Designed Mobile view version of Use Cases
Designed Mobile view version of Use Cases File View.


Tyrepo1 (changes requested):
You made 2 pull requests with the same content. Please remove one of them

PR #789, opened 2024-09-10
Licensing page: context update and background color change

s217034263:
Merge please


Tyrepo1 (changes requested):
Who told you to change the header and footer? Somebody else already finished redesigning them. Your task was to change the home page so it looks like the design
image
Also there are some white lines at the bottom of the licensing page. Please fix that
image

PR #791, opened 2024-09-10
Readding mapping and directions

luilding:


indigovn:
Hi Logan, the code looks good. However, there are a few things might need to address.

1) There is no code to download the dataset from the live source. You can use my method it reads the dataset directly from the live URL. Usage: stops_df, stop_times_df, routes_df, trips_df, calendar_df = GTFSUtils.load_data(url, dataset_folder)

2) The generated map should not save in the execution folder where it has the code in there.
Refer to my one, I have a dedicated folder name maps under the downloaded dataset folder to keep the generated map there.

indigovn (approved):
Hi Logan, the code looks good. However, there are a few things might need to address.

There is no code to download the dataset from the live source. You can use my method it reads the dataset directly from the live URL. Usage: stops_df, stop_times_df, routes_df, trips_df, calendar_df = GTFSUtils.load_data(url, dataset_folder)

The generated map should not save in the execution folder where it has the code in there.
Refer to my one, I have a dedicated folder name maps under the downloaded dataset folder to keep the generated map there.

PR #793, opened 2024-09-11
Licensing Page: context update and background color change

s217034263:


Tyrepo1 (changes requested):
Fix the whiteline without it affecting the layout of other pages

Tyrepo1 (approved):
Looks good

PR #794, opened 2024-09-11
Nauman VC

abidnauman:
Only VC work it


lkocon69 (approved):
Pretty good. Few changes we will make after merge:
- Following images aren't triple road trains and will be moved into applicable categories: image20, image21, image94, image119; img9 in val is standout for having more trailers than a triple road train but we have no category
- label file is incorrectly named "lables" and "label" instead of "labels", will change in separate merge during consolidation

abidnauman:
I have updated and this is the pull request link
https://github.com/Chameleon-company/MOP-Code/pull/825#issue-2525151361

PR #800, opened 2024-09-11
About Page Updated Design v2

koiCoding:


Tyrepo1 (changes requested):
Images are way too big
Why are there black borders for the image?
Color scheme doesn't match the Figma design
Here's the Figma design for reference
image

koiCoding:
Sorry, I was mistaken since there are both wireframe and mobile view in the Figma document that Muhammad sent in the team.
I will update it. Also do you have the exact proportion or how big do you expect the image to appear on mobile devices and on desktop since I can't find it anywhere.

PR #801, opened 2024-09-11
WD_Frontend_Chaya#PR7

chaya221071557:
Login Page Multi-language functionality.


chaya221071557:
I added the header to the login page and used the language button from the header to implement the multi-language functionality.

Tyrepo1 (changes requested):
Can you remove the "no account yet? Sign up" 
We already have a Signup button in the header. Wouldn't make sense to add another one in the body

chaya221071557:
Okay.

PR #802, opened 2024-09-12
Jing Kang - Six Axle Articulated

lkocon69:
Need to include this work to the project


lkocon69:
Work needed to fix the project:
- labels are zero-indexed so they need the correct index of 8
- sample and label files need renaming for condensing
- training sample 71 is a short + trailer/class 2/index 1
- training sample 77 is unknown/class 13/index 12
- training sample 104 is hvt/class 10/index 9
- training sample 117 is 5-axle art/class 8/index 7
- training sample 129 is unknown/class 13/index 12
- train folder is named "training", need to adjust

Kevinbui16 (commented):
Everything is good except the index of both training and validation is incorrect. The class index should be in class 8 not class 0

lkocon69:
I've message Kevin on Friday about approving this request despite some problems as I said I would use this work and fix it. That hasn't happened, so I'm merging without review to bring into my branch "kocon-withmodel' to fix it

PR #816, opened 2024-09-12
Home page color change

s217034263:


Tyrepo1 (changes requested):
Remove the black border and then it should be good

Tyrepo1 (approved):
Looks good

PR #819, opened 2024-09-12
Create HB with chatbot updated output.py

toluvibe:
This is the updated version of my previous code and it has been integrated with chatbot. It can predict health behavior based on the health factor given in Case 5 project and can compare result across three locations.


indigovn (approved):
A good practice is to store the dataset using relative path or environment variable to ensure it works for everyone for instance
current_directory = os.getcwd()

Don't use this file_path = r"C:\Users\akind\OneDrive\Desktop\782 work\HB1.csv"

PR #834, opened 2024-09-14
MOP_WD_Adrian_Weerasinghe_PR6 [update]

adrianweerasinghe97:
Things that have been added

1. Updated Cypress. updated "cypress.config.ts" by adding "baseUrl and test file patterns"
2. Created a Cypress test for the Chatbot component
3. updated the jenkinsfile with the cypress test stage
4. updated the babel file


Tyrepo1 (approved):
Looks good

PR #850, opened 2024-09-16
predicting CO2 emission by day/month and year

MobassherNomani:


luilding:
Going to playground instead of AI folder, please change to that

PR #858, opened 2024-09-16
WD_FRONTEND_THAMASHA #PR3

thamashag:
**Updated signup functionality to use 'mop' database in MongoDB Atlas**

- Adjusted the MongoDB connection to ensure user data is stored in the 'mop' database
- Updated the environment variables to include the correct database name in the connection string.
- Refined the backend logic to connect properly to the 'mop' database for storing user information.
- Verified that the signup process now correctly stores user data in the intended database.
- Ensured data consistency and improved backend functionality by targeting the correct database.


Tyrepo1 (approved):
Looks good

PR #872, opened 2024-09-17
Updated VC

abidnauman:


lkocon69 (approved):
Pretty good. I'll need to clean up the index change to 11, because that will be the actual index of the triple road trains

PR #896, opened 2024-09-18
Data Science Team - Soil Analysis and Prediction - 100% complete

trostov:
Use case includes:
- Cleaning data to ensure it is standardised and consistent
- Separating and merging dataframes 
- General data wrangling to ensure the datasets are easier to work with
- Multiple visualizations comparing different variables
- Analysing these visualizations, seeing how the variables impact the results
- Initial machine learning to see how hyperparameter and model selection impacts the results
- Encoding two different categorical variables into numerical ones 
- Selecting and further tuning a final machine learning model
- Encoding both categorical variables at the same time, and training a final model on the full dataset
- Inputting custom predictions to the final model
- Visualizing these results with colour coded map markers using Folium


kitkatch (approved):
Hi Thomas, 

Awesome work, I love how you have explained what you are doing so clearly, very educational.  Do you think this use case would benefit with a conclusion?  It's up to you obviously. :)

HarshDW09 (approved):
Hi @trostov  , 

Your use case is well-constructed. I agree with Katrine that including a conclusion or final analysis would help users better understand the main goal of the use case.

lynzaa (commented):
Hi Thomas

Good work! A conclusion report would be an add-on but it's definitely upto you.

Thank you
Liny Alias

PR #901, opened 2024-09-18
Project handover

sinankilci0:
- Additional data wrangling
- Combine datasets
- Analysis and report
- Tidy ipynb
- Create html and json


trostov (approved):
Very nice usecase, good complete writeup with sufficient comments and lots of visualizations . Please change the usecase duration though, its currently set as "6040 minutes". Some of the graphs potentially look a little cluttered, but this is ok if you believe it was the best way to visualize the content.

Samiha-Haque (changes requested):
@sinankilci0 Amazing usecase! The files look good too.
However a few minor changes are required:
1. Both your maps need legends and a title -> After making the changes I think you need to update the html file as well and use the upload one instead.

trostov (approved):

Samiha-Haque (approved):
@sinankilci0 looks good! Approved.

PR #939, opened 2024-09-21
Data Science - Evaluating Public Transport Accessibility to Places of Interest in Melbourne - 100% Usecase Completed

emmanuel2199:
Updates added:
- I worked on future project implementations.
- Integrated a new dataset 
- Created visual representations of the public transport infrastructure in Melbourne.


VaishaliThakoor (changes requested):
Hey Emmanuel, your use case looks great! A few things I noted were:
1. There are a few blank code lines, it would be better if you delete those so that it looks clean!
2. The graphs at the beginning such as distribution of landmarks and distribution of tram stops do not have any descriptions, it would be good if you can add a few lines to describe what the graph is about and what it has achieved. Make sure all your plots and graphs have descriptions!
3. When you performed spatial analysis, you saved the heatmaps as HTML files and those are displaying an error as the result of that code when I ran the ipynb file, do you especially need to save them as HTML in your analysis? If not can you  just display the heatmaps in the code itself instead of saving them as HTML first. **@Samiha-Haque might have to check that and tell us if she can see the error too**
4.There is a placeholder file in your folder that you will need to delete.

Let me know if anything is not clear! Otherwise great job!!

Samiha-Haque (changes requested):
Hi Emmanuel, the use case looks good. However, a few modifications are required:
1. Removed your folder from Playground.
2. Removed placeholder.txt
3. Import all the libraries and packages at the beginning of the code file instead of in between code blocks.
4. In line 23, it shows and error because it is missing Accessibility_Heatmap.html. If you want to add it then keep Accessibility_Heatmap.html in your use case folder, otherwise remove that code block
5. Add legends and titles to your maps.

emmanuel2199:
Hi @VaishaliThakoor and @Samiha-Haque, I have made the following changes to the folder and files:
- Deleted the placeholder.txt file.
- Added descriptions to the visualizations
-  Removed the Accessibility_Heatmap.html and directly implemented it in the code
- Imported all the libraries and packages at the beginning of the code file
- Removed the folder from Playground

VaishaliThakoor (approved):
Great job!  @emmanuel2199

Samiha-Haque:
Hi @emmanuel2199 titles and legends to the maps would be great! but otherwise I think it's good to be merged

Samiha-Haque (approved):

PR #980, opened 2024-11-18
DATA SCIENCE | Moving webapp directory into .ARCHIVE directory - Not Used

Supratim-S220205268:
This is a cleanup task assigned to me by mentor Mollie Fernandez. A few of the directories have not been in use for the past few trimesters and need to be moved to the ".ARCHIVE" folder.

I am only moving one directory - "webapp" to the ".ARCHIVE" folder for now. Once this is approved and merged successfully, I will be moving the other directories as well.

Folders affected by this change:
- MOP-Code/webapp  
-  MOP-Code/.ARCHIVE


molliefernandez-mentor (changes requested):
Everything looks great!
The only small change I would request is that it looks like you have created an additional 'webapp' folder unnecessarily. See image for reference.
webapp

molliefernandez-mentor (approved):
Thanks Supratim, all looks good. Please merge and delete your branch.

PR #1006, opened 2024-11-22
Kheang

KheangDeakin:
Adding in my usecase Environmental Resource Consumption and Service Demand in Melbourne


Sabihulhassan98 (changes requested):
Hi Kheang, Great work with the use case but I noticed that you have two .ipynb files in your branch. Can please delete the second file called Untitled.ipynb.
Also can you please follow the pull request documentation and add a proper title and description while making a pull request

PR #1031, opened 2024-12-05
yolo work with sahi

Fragmentszz:
I’ve implemented code that integrates YOLO with the sahi technique. <font color="red">**This includes the implementation of multi-scale SAHI detection and an ensemble learning approach (to be confirmed) for merging labels.**</font>
image
While not all items are detectable at this stage, I believe that the multi-scale approach is a significant advancement. I’m looking forward to your feedback and suggestions for further improvements.


molliefernandez-mentor:
Old Pull Request

PR #1033, opened 2024-12-06
Added new way of working with PTV data using APIs version 3

indigovn:
1. Added new way of working with PTV data using latest PTV APIs.
2. Added 3 services disruption for train, tram and bus have been implemented in this checkin.
3. Update landing page by added more commands for Tram and Bus


luilding (approved):
Hey Alex, good work. However this is a branch conflict in the domain.yml that needs to be fixed

PR #1087, opened 2024-12-14
Updated usecases to get from ready to publish and implemented carousel functionality to usecase in homepage

MAXBASH:
Updated usecases to get from ready to publish and implemented carousel functionality to usecase in homepage


BenTran203 (approved):
Good work updating the publish function and implemented carousel functionality.

PR #1106, opened 2025-01-12
Data Science- UC00139_Optimization_of_Outdoor_Activities- 100%

ananyaKrishnan-2023:
Final week push:
All steps completed:
1. Libraries and packages are installed. Besides, data from various packeges are imported.
2. Data procuring or acquisition where data is gathered from APIs provided by the company. Moreover, the data is showcased in entirety for obtaining summary of primary and statistical infrmation.
3. Data cleaning and pre-processing, here data is organised such that the precise details are forwarded for any sort of processing or computation. Next, latitude and longitude values extracted from given location co-ordinates. Followed by calculation of foot-traffic and geo-spatial analysis is conducted for each dataset. Further, the determined geometry column of the dataset is activated. Then, every datasets are combined spacially.
4. Exploratory data analysis part is where vital characteristics are encapsulated, patterns are discovered, anomalies spotted, and data is visualized for further investigation. The cases at the moment are pedestrian flow in open-space (Melbourne city) with and without comfort-resources availability and accessibility.
5. K-means clustering is applied to get count of active-individuals of Melbourne city.
6. Visual illustration of in-depth analysis of pedestrian traffic in Melbourne city.


Supratim-S220205268 (approved):
Hey Ananya! Just checked out your use case, and the work looks really great. I see you have done a lot of data wrangling to achieve your goal. My suggestion is to uplift your work using more interactive visualizations. Apart from that, great work. Both json and html files also look okay!

Cheers :)

PR #1117, opened 2025-04-03
Data science_ New use case_T1 2025_UC00153_Biodiversity_corridor_creation_35% completion

Jnaneshwari-beerappa:
1. Successfully imported 3 spatial datasets using a dynamic API-based function.
2. Converted geo_shape into proper geometries using shapely and GeoPandas.
3. Filtered high-potential laneways using fores_rank == 1 to focus on top candidates.
4. Performed spatial join between laneways and canopies to identify connected segments.
5. Ensured all datasets share the same CRS (EPSG:4326) for accurate spatial analysis.
6. Created a clean and dynamic Folium map with layers for laneways, planting zones, and canopies.
7. Included detailed popups for area, greening type, and schedule in the interactive map using follium.
8. Updated progress in MS planner
9. Mentioned use case and labelled use case as UC00153_Biodiversity_corridor_creation
10. Regularly updating my tutor about progress and seeking guidance.


SaiPriyamvada21 (changes requested):
hi jnaneshwari
1.can u please use the fixed template 
2.please check your files as well you uploaded 2 similar files 
3.please check the cell at the start of ur use case

PR #1138, opened 2025-04-04
DATA SCIENCE -UC00157_Night_Time_Safety_index-30% complete

ajaiysentil:
Week 5 pull request for review from Sprint 1 

Use case includes: 
-Scenario
-Use case introduction 
- API v2.1 integration  
-Data import and Cleaning 
-Filtering and Preprocessing 
- Initial geospatial visualisation


Nish245 (commented):
Hii Adithya 

The use case scenario is detailed and explained well. API key is visible while using v2.1. I would suggest you add some different titles for each subtopic you start. Also avoid using paragraphs for explanation. It will seem boring for few people. Add pointers for explanations. Use BOLD wherever necessary to highlight. 

Apart from maps, also you use some other plots, you can use seaborn library or matplotlib whichever feels comfortable and suitable for the visualizations. Moreover, add your naming convention into the starting of the use case in a markdown chunk.

Nish245 (approved):
Thank you for the changes

PR #1146, opened 2025-04-05
DATA SCIENCE UC00148_CommunityWellbeing_TrendAnalysis -50% Complete

Nidhi-kanchepalle:
Week5 Pull request for review from Sprint1
Use case includes:
-use case introduction 
-Import Data using API 
-Handling of missing values 
-Exploratory Data Analysis
 -Analysing specific trends (eg: smoking behaviour) 
-Suggestion to improve intiatives based on trends


Supratim-S220205268 (approved):
Hello Nidhi! You are doing an amazing job justifying this use case. I really like how you've structured your use case and how you've started your analyses, including some insightful visualizations like that Correlation Heatmap. Overall, it's great work! All checks are in place, so all good.

My feedback to you, is to work on these points to take your use case towards perfection:
1. I understand your use case is based on analysing the CoMSIS survey data and the opportunities are endless. I see you are analysing various factors like Smoking and Vaping Trends, and Aboriginal Knowledge. I would suggest you focus your use case around a specific objective. That would give the reader a more definitive interest in going through this document. For example, if you are analysing smoking trends, try to look for avenues around it, so you can base your use case on it. Again, this is only a suggestion, it's totally up to you how you want to take this forward.

2. Try to include a paragraph before each code chunk, explaining what you are about to do, and why do you need to do it? This will make it look more like an actual tutorial for someone who has no background in data science. You have added the explanation for Handling Missing Values. Similarly, it would be better if you add an explanation for Inspecting the datasets, Deleting Unwanted Columns, and Exploratory Data Analysis.

3. You have used both American and Australian English in your use case. For example, you have used both words - 'analysing' and 'analyzing' in your use case. Make it follow just the Australian format.

vivekolladapu (approved):
Hi Nidhi, everything looks really good, the format is followed, the code is well-organized, it’s easy to follow, and the visualizations make the insights easy to understand. The only small thing I noticed is that there are a couple of spelling mistakes in the response values, like “daliy” instead of “daily” and “occassionally” instead of “occasionally”, I would suggest correcting these typos using a string replacement, and with that, the use case will be even more polished.

PR #1170, opened 2025-04-16
WD_Backend_Sushma#LoginSigninApiNavigationAndProfileBasedloginPage

223743838:
Created the login route and resolved the issue which we were getting while signIn and login page . Also Updated the login page after validating the new user with user name and Logout button on header.


DanishKumar1001:
Fixed formatting issues and conflicts. Merged to main

PR #1178, opened 2025-04-28
WD_FrontEnd_HemakshKatal #Redesign_LicensingPage_LightMode

softwareHe:
Redesigning the licensing page(Light Mode): 
- Removing the previous content wrappers that justified the content left and right both and then justified the whole content to the left only
- Added the icon for the chatbot 
- Made the content responsive by using flex wrapper
- Made necessary margin gaps that were missing earlier between written content 
- Each content paragraph has been separately containerized 

[BEFORE AND AFTER LICENSING PAGE.pdf](https://github.com/user-attachments/files/19934281/BEFORE.AND.AFTER.LICENSING.PAGE.pdf)


DanishKumar1001 (approved):
License Page Lightmode approved

PR #1226, opened 2025-05-14
Environmental sustainability moving usecases

HarshDW09:
Moved all the use cases from year 2024 and 2023 into Environmental_Sustainability folder


molliefernandez-mentor (approved):
Looks great, Thanks Harsh

PR #1301, opened 2025-05-17
Add mental health chatbot notebooks

Benard-Kimutai:
Pushed 3 notebooks related to chatbot, fine-tuning and training.


faizuuuuuu (changes requested):
Thanks for your work on the chatbot! 🙌 Just wanted to flag something important before i proceed with the pull request.
 your files were added directly to the root MOP-Code/ directory, instead of inside the structured path we’re all using:
swiftCopyEditartificial-intelligence/T1_2025/T1_2025_LLM_Chatbot/your_project_folder/ 
🔄 What needs to be fixed:
To keep things clean and consistent across the project:
Please create a subfolder for your chatbot inside this directory:
 
For example:
gemini_mental_health_bot_yourname/
Move your files (e.g., .ipynb, .py, .sh) into that folder.
From your branch, run:
git add .
git commit -m "Moved chatbot to correct subfolder"git push origin your-branch-name
After that push your code to your branch.
verify that the files have been deleted. also check if the files have been added to the new subfolder
then the pull request would be automatically updated

PR #1312, opened 2025-05-18
DS Team- UC00139 Outdoor Activities Optimisation- 100%

ananyaKrishnan-2023:
Use case tutorial completed. Prepared for the final check.
Background
Code Implementation
Conclusion


Supratim-S220205268 (approved):
Hello Ananya!

I think you've done great work here! I like your use case idea, it is unique in context of other use cases this trimester.

You're following the template correctly and using the API appropriately as well. You've added some very interesting visualizations and explained the workings as you go step by step. You've put together a lot of work in this report.

Having said that, I do believe that you could have taken this work to the next level by making it a little more visually appealing.

Here are my Suggestions for future:
1. Removing the pip statements. The packages are already installed, you don't need it anymore. It'll make the report look more neat.
2. Provide context before each step. Discuss Why and What is happening?
3. Discuss the Observations and Findings with regards to the visualisation outputs.
4. Add Recommendations.

HarshDW09 (approved):
Looks good!

For future reports, you could structure them more like formal reports, providing context on why they are being created, outlining your design decisions, and perhaps including recommendations as well.


Meets all the requirements.

PR #1313, opened 2025-05-18
Added files for Fine_tuning, training and UX/UI

Benard-Kimutai:
Files added were fine_tuning script, training files and ux/ui created using gradio.


faizuuuuuu (approved):
Mistral 7B Code in the right folder.

PR #1331, opened 2025-05-18
Naga nikhil woopalanchi

nikhilnaga1:
Updated the json and html files, use case ready to be moved to focus group


vivekolladapu (approved):
Hi Nikhil,

Everything looks good.

thanks!!

PR #1345, opened 2025-05-19
Rathanak

Rnak2:
Pull request for use case 10, CNN and LSTM training files


rickspacecar (changes requested):
Hi Rathanak, great job getting your UC10 implementation in and organizing the folder structure correctly.

Just two important issues flagged by GitHub’s code scanning that need to be fixed before I can approve:

Filename Sanitization
Please use secure_filename() from werkzeug.utils to prevent path traversal risks.
Copilot has already suggested the correct fix.

Flask Debug Mode
Avoid running the Flask app in debug=True by default. Use an environment variable instead, as suggested, to keep your app production-safe.

You can commit the Copilot suggestions directly or push your own fix — then let me know and I’ll approve right away. Thanks!

PR #1351, opened 2025-05-20
Move use cases to Education & Teaching group folder

jennynguyen2112:


molliefernandez-mentor (changes requested):
Thanks Jenny, Can you please sort them into year, with the small amount of use cases you could just have them sorted into years, don't worry about the trimester.

molliefernandez-mentor (approved):
Thanks, ready to merge

PR #1411, opened 2025-07-25
WD_FrontEnd_Pratik_Khadka#ConsistencyBetweenButtons

pratikdka:
Added the consistency between the button when the dark mode is enabled. The same hover effect is added now between them.

**Before Changes - Inconsistency can be seen**
image

**After changes**
image


molliefernandez-mentor:
old pull request

PR #1437, opened 2025-08-08
DATA SCIENCE TEAM - Active_Transport_and_Air_Quality_Impact - 50% complete

nnbson:
Week 5 pull requests for review from Sprint 1

Use case includes: 
Import data using API v2.1 GET requests with /exports extension on City of Melbourne Open Data
Import data using read on data from Data Vic
Create a function to integrate zip files from Transport Victoria
Filter of dataframes
Merge similar datasets with different periods
Reduce data dimensionality to only the features that are needed
Convert data types to the appropriate types
Visualise the relationship between the response and explanatory variables
Find the correlation between the response and explanatory variables


PJRyn (changes requested):
Hi Shawn, I like how your use case is developing! Also, that you have taken on the challenge of working with a pretty diverse set of data and making it work. Great job on getting to the visualisations especially since ones like the “Trend of pedestrian/bike usage and air pollutants in 2022 - 2023” is far from simple. But there are a few things that need to be fixed up before it is ready to be merged.

First, you have got the file location right, but the name of the file needs to be the name of your use case (‘UC00175_Active_Transport_and_Air_Quality_Impact’). Since when it gets published into the Environmental Sustainability T2/2025 folder we need to know what file we are looking at. 

Also, it is totally fine to use non-MOP data. But when you do you need to include links to the dataset in the file as well as adding a folder into the dependencies folder: https://github.com/Chameleon-company/MOP-Code/tree/master/datascience/usecases/DEPENDENCIES 
Just add a folder called ‘UC00175_Active_Transport_and_Air_Quality_Impact’ and upload the files into that. That way users can recreate the use case on their own machines when it is published. It also means that if the non-MOP sources are removed or have their locations changed people within the organisation can find your data. 

The biggest change is that you need to follow the use case template, it will guide you through tings like writing a title, introduction and guide the reader through your use case. You can find it here:
https://github.com/Chameleon-company/MOP-Code/blob/master/datascience/usecases/usecase_TEMPLATE.ipynb


Also make sure you add in more information to what you are doing and why. Some of the code comments are fine, but we are also describing this for a non-technical audience of readers. So adding some markdown sections to explain why you are importing data the way you are or making your visualisations as well as explaining what you found from them. It might also help to break up your file into sections e.g “data imports”, “data wrangling” or having specified sections for each part of your analysis.

And one really minor one is that some of the code comments weren't consistent when going through the ‘pedestrian_count’ data cleaning and concatenating. Like how  `pedestrian_count_2023_2025_reduced`  was missing step 5 and `pedestrian_count_2022_reduced` was missing step 2. Also you re-imported seaborn in your correlation matrix, while that usually doesn't cause any issues in Python it is best practice to call them once when you do all of your imports.

Overall, it was super promising and the only things you really needed to fix was just documentation related stuff. So it’s a great start! Just fix up those bits and you should be good to go!

Sam-Sep1ol (commented):
## Functionality
- Notebook runs end‑to‑end and produces the expected visualisations from the loaded data.
- Recommend a brief opening cell outlining purpose, data sources, and any assumptions, plus a setup cell listing required packages and versions to aid reproducibility.

## Readability
- Plots are easy to read and align with the use‑case goals.
- A few names are terse/generic; using self‑describing names (e.g. stops_filtered, daily_counts) would make intent clearer.
- Add short markdown notes before major steps to explain the rationale (why a filter/merge/aggregation is applied), not just the mechanics.
- Consider grouping imports in the first cell and keeping long cells bite‑sized for easier scanning.

## Positives
- Workflow is coherent: data load → cleaning → transformation → visualisation.
- Visualisations are relevant and support the narrative of the analysis.
- Code is mostly consistent in style and structure, making it straightforward to follow.

PR #1440, opened 2025-08-08
UC00179_School_Enrolment_Trends - Sprint 1

pooyafo:


MuzamilRafiq (approved):
The notebook presents enrolment trends clearly with well-labeled visuals. A useful improvement would be to add short markdown explanations before each plot to highlight the key insight being shown. A brief summary at the end would also help tie everything together.

PR #1447, opened 2025-08-08
DATA SCIENCE TEAM | UC00172_Urban_Biodiversity_Atlas_and_Habitat_connectivity | 45% Complete

vamshik27:
Use case: UC00172_Urban_Biodiversity_Atlas_and_Habitat_connectivity
Focus Area: Environmental Sustainbility
Submission for Sprint 1:
Accessing the Dataset with API Data Cleaning and Handling of Missing Values Exploratory Data Analysis Geospatial Analysis.


PJRyn (approved):
Hi Vamshi, this use case is fantastic. It followed the correct location and it was correctly named. The dataset used the API call correctly. It also followed the template to a tee. Every  section is clearly labeled and has just a clear attention to detail and the user experience is also clearly accounted for. Things like adding tables when showing the datasets used or when filling in the missing taxonomic information in the insect records, you not only explained what you were doing in the markdown but also clearly in the comments of the code. Honestly, it made me want to improve my use case! 

The only feedback I can give, and this is such a small criticism, you could expand on your analysis and breakdown of each visualisation. They are really quick and concise but maybe giving a bit more of your thoughts would be welcome. Honestly though that is more of a stylistic criticism and you don't need to incorporate that at all. 

Fantastic work and it was really cool to see the polish and attention to detail. I cannot wait to see the completed use case, really amazing!

PR #1480, opened 2025-08-10
DB Design

vossiepkv:
Created DB ERD and DFD as well as finalised DB Schema


vossiepkv:
Added Db Design PDF's

PR #1506, opened 2025-08-25
DATA SCIENCE TEAM - Active_Transport_and_Air_Quality_Impact - 60% complete

nnbson:
Week 5 pull requests for review from Sprint 1

Use case includes:
Change the file name
Add non-MOP dataset to DEPENDENCIES
Adjust the use case notebook to align with use case template
Import data using API v2.1 GET requests with /exports extension on City of Melbourne Open Data
Import data using read on data from Data Vic
Create a function to integrate zip files from Transport Victoria
Filter of dataframes
Merge similar datasets with different periods
Reduce data dimensionality to only the features that are needed
Convert data types to the appropriate types
Visualise the relationship between the response and explanatory variables
Find the correlation between the response and explanatory variables

### DIFF
--- a/datascience/usecases/DEPENDENCIES/UC00175_Active_Transport_and_Air_Quality_Impact/Link to all the non-MOP dataset.txt
+++ b/datascience/usecases/DEPENDENCIES/UC00175_Active_Transport_and_Air_Quality_Impact/Link to all the non-MOP dataset.txt
@@ -1,2 +1,0 @@
-All dataset can be found in the Google Drive link here, since the files are too big to be uploaded on GitHub:
-https://drive.google.com/drive/folders/1kB62So68Y6jGd7kcLSHDI476-jyHKRA5?usp=drive_link


PJRyn (changes requested):
Hey! Sorry it took me a couple days to get to this. There are massive improvements in this git pull request. The use case follows the template well and you have done a good job with the introduction and explanation. It's going in a fantastic direction! 

The only real issue is that the external ZIP files have made your life a bit harder here since you did put the “Link to all the non-MOP dataset.txt” in the correct location and that is great! But it does have a Google drive link that is private. But I don't think we can use Gdrive links for our storage since the goal is to have data after we have left Chameleon (and maybe Deakin). So your data will need to be stored in a location that the future Chameleon members (and Melbourne city council after this program is completed) can access them. I do understand that you might be throttled by upload size. The online client has a size limit of 25MB which is pretty constricting. But the terminal based git client has an upload size of 100MB. You might need to split your data up into parts and upload them that way. If you have any issues or are stuck it might be worth mentioning the issue to Mollie. 

Otherwise the only other feedback I have, and this is far less important. So don't sweat it that much but the use case did leave the reader without much explanation of why things were being done. For example when working with the “Pedestrian_Counting_System_Monthly_counts_per_hour_2009_2022.csv” you had used chunk sizes to break up the data. While for a technical audience this makes sense after reading the code. But this might be read by a non-technical audience who might need the explanation to keep up. Same for the visualisations, they just need a few sentences here and there to explain your takeaways and conclusions. But this is just a tiny thing so don't worry about implementing it.

The work looks really good and high quality. And it does fit the requirements. The code and visualisations are also super solid. It's clear you did put a lot of time and thought into it. And you have given yourself a real challenge with this use case. To be quite honest I don't think I have seen any other use case handle this much data and use so much data from other sources. I really don't want this to discourage you since your work is great. I would recommend mentioning to Mollie the size and scale of your use case. Since the technical contribution, at least in my opinion, is really great. And due to the complexity you might be able to work on it as a senior. Although I would totally understand if you wanted to get this one done this trimester! Just work on having a permanent storage solution for your datasets and I think you will be in the clear to merge.

nnbson (commented):
I have added a notebook of my use case to the Playground folder and added all the datasets to DEPENDENCIES folder.

PR #1548, opened 2025-09-15
Add functions for travel time calculation

jennynguyen2112:


ndaccong (approved):
The chatbot can work smoothly now. It is still missing the distance calculation part, which is expected to be added in the next commits.

PR #1573, opened 2025-09-18
UC00172_Urban_Biodiversity_Atlas_and_Habitat_connectivity-Ready to Publish

vamshik27:
Moving the completed and reviewed use case to "Ready to Publish" folder
Delete folder from Playground


PJRyn (approved):
All the folders and the use case are in the right place. Seems like it is all good to merge in my opinion!

PR #1576, opened 2025-09-19
DATA SCIENCE TEAM - Active_Transport_and_Air_Quality_Impact - Ready to Publish

nnbson:
Use case includes:
- Use case introduction
- Use case scenario
- Use case background
- Datasets
- Packages
- Import data using API v2.1 GET requests with /exports extension on City of Melbourne Open Data
- Import data using read on data from Data Vic
- Data Preprocessing and Wrangling
- Data Analysis
- Data Visualisations
- Model development
- Use case results
- Recommendations for further research


PJRyn (approved):
Looks good! All three needed files are here and they are in the correct location! Well done on this use case. It was awesome to see its development and thank you for you hard work

Sam-Sep1ol (commented):
Awesome work Shawn, looks great.

PR #1619, opened 2025-09-24
UC00189 Weekend Business Activity Zones use case files

sahithipinnam:


aneeshpedram1:
Hi Sahithi, please rename the html version of the use case notebook for consistency. There is an extra '-3' at the end.

Also, you've created an extra docs folder in the MOP-Code repository, which is not required. Please add any extra documents that you may need for your use case into the ready to publish folder of your use case itself.

sahithipinnam:
Hi Aneesh, I have made the required changes. Please review and let me know if anything else needs adjustment.

aneeshpedram1 (approved):

PR #1645, opened 2025-09-30
UC5: relocate files into T2_2025 mentor-specified path

ramadevimiriyala:


ndaccong (approved):
The folder is now in the correct path. Good to go.

PR #1651, opened 2025-10-02
Chore/add editorconfig

rohitbondala277:
Adds a root .editorconfig to keep line endings, indents, and whitespace consistent across editors.


molliefernandez-mentor:
not relevant

PR #1652, opened 2025-10-02
Chore/add gitattributes

rohitbondala277:
Normalizes line endings to LF and marks common binary/text types to avoid cross-OS issues.


molliefernandez-mentor:
Not Relevant

PR #1716, opened 2026-03-31
DS - UC00206 - Added a new data source folder in Dependencies folder

arvindronin:
Added new folder UC00206_School Enrolment trends
Included 4 CSV files containing enrolment data from 2022 to 2025
Data sourced from internet for trend analysis


manya0033 (changes requested):
Hey Arvind, I've gone through the datasets on your branch, and they all look cleaned and ready to go.

One issue though, the PR is showing Files changed: 0, and I've checked that the files aren't on master either. What this means is that the 1 commit in this PR doesn't actually contain the dataset files. The files exist on your branch (I can see them when I browse it), but they were probably added in a different commit that isn't part of this pull request. If we merge this as-is, master won't get the files, only the empty commit gets applied.

Could you try re-adding and pushing the files, so they show up in this PR? Or if that doesn't work, it might be easier to close this PR and open a new one that includes the commit with the actual files.

Happy to approve once the files changed tab is showing the 4 datasets.

PR #1721, opened 2026-04-02
DATA SCIENCE | UC00234 Urban Forest Risk Assessment | Address Review Feedback

aidanuni:
Sprint 1 (data pipeline) and Sprint 2 (ML model) are now complete.

Changes since last review:
- Fixed CRS from EPSG:7844 to EPSG:7855 — sensor distances now in meters
- Consolidated 5 notebooks into single pipeline notebook
- Added DEPENDENCIES/ folder and README
- Completed ML model: Logistic Regression baseline + XGBoost with class imbalance handling
- Added SHAP explainability — per-tree feature attributions
- Layer 3 (RAG/LLM) deferred to next trimester


manya0033 (changes requested):
Hey @aidanuni  thanks for walking me through Sprint 1 - the structure is clean and I can see a lot of thought went into the feature engineering (rolling temp averages, heatwave flags, consecutive hot days is a nice touch). Before I approve, a few things to address:

The CRS warning in 03_spatial_joins.ipynb is still showing because EPSG:7844 is actually a geographic CRS (GDA2020 lat/lon), not a projected one. You can see the effect in the distance stats,  the column is labelled sensor_distance_m but the values range from 0.00002 to 0.05, which can't be metres for 82k trees. They're actually still in degrees. Could you switch to a projected CRS like EPSG:7855 (GDA2020 / MGA zone 55) before the sjoin_nearest calls? That'll give you genuine metre distances and the warning will go away.
The notebooks reference local paths like ../data/processed/feature_table.csv but I can't see where the raw data is actually coming from. If you're pulling the datasets from the Melbourne Open Data portal via API v2.1, could you include those API calls directly in the notebook so reviewers can reproduce the pipeline end-to-end? Otherwise, if any of the data is external, a CSV version should go in the DEPENDENCIES folder as per the checklist.
One suggestion, would it be possible to consolidate everything into a single notebook rather than five separate ones? Our use cases are meant to read as step-by-step tutorials, and having the whole pipeline (exploration -> cleaning -> spatial joins ->feature engineering -> ML model) in one notebook with clear markdown headers between sections would make it much easier for readers to follow along and reproduce. It also avoids the issue of needing intermediate files saved to disk between notebooks.
A short README in your project folder would also help- just a few lines on the pipeline order, where the raw data comes from, and any setup notes.

No worries that the ML section is mostly empty, the PR title is clear that Sprint 2 is 20% in progress. Happy to re-review once those are sorted!

manya0033 (changes requested):
Hey Aidan, massive improvement from the last review - you've addressed everything I flagged. The CRS fix is properly documented, the single notebook flows really well, the README is excellent, and the XGBoost + SHAP work is genuinely impressive. A few final things before I approve:

Your .gitignore excludes the DEPENDENCIES/ folder, so the raw data files aren't in the PR. Reviewers would need to manually download 6 files from different sources to reproduce anything. Could you either remove DEPENDENCIES from the gitignore and commit the data files, or upload them to the repo-level datascience/usecases/DEPENDENCIES/ folder?
Could you add the standard use case template header at the top (usecase-title, authored-by, duration, level, pre-requisite skills divs)?
It would be great to add a few markdown interpretation cells after the main outputs - especially after the EDA stats, the spatial join distances, and the model results. Just a sentence or two explaining what the numbers mean and why they matter.
These are minor - the technical work is solid. Happy to approve once those are in!

aidanuni:
@manya0033 — all three items addressed in this latest commit. Ready for re-review whenever you have time. Thanks for the detailed feedback!

molliefernandez-mentor (approved):

molliefernandez-mentor:
Use Case looks good Aidan, Once you get a re-review from the others it's good to merge.

manya0033 (approved):
All three items addressed - DEPENDENCIES folder committed with all 6 data files, use case template header added, and interpretation cells included. Mollie has approved. Great work on this Aidan, approving!

PR #1733, opened 2026-04-03
DATA SCIENCE UC00211_PEDESTRIAN_CONGESTION_USING_TIME_SERIES_ANALYSIS

diy-rob:


vysh-afk (approved):
Great work diyona,keep going

manya0033 (changes requested):
Hey @diy-rob , nice work on this! The structure is clean and I like that you included a baseline model to compare against. Just a few things before I can approve:

The dataset is being loaded via the direct CSV download URL- could you switch it to API v2.1 (the /records endpoint) instead, as per the PR checklist? If it's an external dataset that can't be pulled through the API, a CSV version should go in the DEPENDENCIES folder.
Small thing- cells 13 and 15 both call pd.read_csv on the same URL, so the first one can be removed.
Could you add a PR description too? Would be great to drop in a few quick "Key Insights" markdown cells after your main plots so readers know what to take away from them.

Let me know once those are sorted and I'll re-review!

manya0033:
Also, The README.md file is empty, could you fill it in with a quick overview of the use case, how to run the notebook, and any dependencies or remove it if it is there as a placeholder?

PR #1877, opened 2026-05-14
Project 8

Sav289:
[Added validation testing and accuracy results of object detection model using a couple of metrics. Was looking to change the scope of the object detection model but it deemed to be too dificult at this stage of the project


hamizhassan16:
Changes look good

PR #1890, opened 2026-05-15
Add gallery lightbox navigation controls

abdullahsher445:
This PR improves the gallery image popup experience by adding previous and next controls inside the lightbox. Users can now move through gallery images directly from the popup without closing it or clicking each image separately.

Changes made:
- Added previous and next buttons inside the image popup
- Added keyboard support using left and right arrow keys
- Added image counter inside the popup
- Kept the existing gallery layout, card design, pagination, and page styling unchanged


abdullahsher445:
This PR updates the navbar so the Profile link is only visible when a user is logged in.

Changes made:
- Added a filtered nav item list based on login status
- Hid the Profile navbar item for logged-out users
- Kept all existing navbar styling, layout, and behaviour unchanged

PR #1892, opened 2026-05-15
DATASCIENCE - UC00229_Smart_Streetlight_Fault_Detection - 100% Complete 

hamizhassan16:
Uploading files to Dependency Folder


manya0033 (changes requested):
Hey Hamiz, a few things before I can approve:

Please add a proper PR description with the reason for the pull, what's included, and a list of changes as per the PR guide.
The PR title should follow the standard format: "DATA SCIENCE | Project 8: Smart Streetlight Fault Detection | % Complete"
The split datasets in DEPENDENCIES look fine structurally but just double-check with Mollie that she's happy with the total size going into the repo since it's several GB across the ExDark and Predicted Streetlights chunks.

Happy to re-review once the description and title are updated!

molliefernandez-mentor (changes requested):
Size of the dependency files are fine provided GitHub will accept them!

I would love to see more explanation at each step. Some of the language you use is fine for someone who knows what they are doing but the audience for this is someone who has never done this before. You need to explain each step, what you are dong and why.

hamizhassan16:
I have added more description to the files as requested by Mollie, have finalised everything and making final push to Ready To Publish folder

hamizhassan16:
now all dependencies uploaded as well awaiting review:)

molliefernandez-mentor (approved):
Thanks guys, please wait for the other review to approve but all good from me

manya0033 (approved):
Thanks for updating the title, description, and adding more explanations as Mollie requested. The structure looks good. One small note for future reference - cell 1 has a hardcoded local path that should ideally be relative so other users can run it. Approving!

PR #2021, opened 2026-07-20
Use case template updated T2 2026

AtishayDeakin:
Minimal Clean template: underline-style section headers instead of solid colour blocks, no numbered headings.


molliefernandez-mentor (changes requested):
Hi Atishay, just a few starting feedback points. coloring and style/formatting is good just want to make it clearer for the quality managers and future students how the use case as a whole should look as well. It definitely needs an expansion of the example section. 

Can you please make it clear that each section needs to be numbered and sub sections double numbered i.e. 
1. heading
1.1. sub heading
1.2.5. sub sub heading etc.

The MOP logo image isn't working and probably doesn't need to be in the use case, it will be on the website. 

Can you add a section which states the date the use case was created as well under Author

Please include a contents page which outlines the numbered content below. - this should be after the key takeaways but before the code content starts.

I want it clear that each heading needs a description underneath it. 

The style guise section should be applied to the content example section, so it is clear how it will looks. 

Still needs a lot of work - some comments for now, I look forward to your updates :)

molliefernandez-mentor (approved):
Looks good, I just made a few minor changes :)

PR #2031, opened 2026-07-23
Fix dark mode text in use case detail page

abdullahsher445:
## Summary

This PR fixes the dark mode text visibility issue on the use case detail page.

## Changes Made

- Updated the notebook/content wrapper to support dark mode text colours.
- Added dark mode child text styling so content rendered inside `NotebookRenderer` does not stay black on a dark background.
- Replaced the old iframe fallback with a styled HTML content wrapper for better dark mode support.
- Improved readability of use case content in both light and dark mode.

## Why This Change Was Needed

Previously, some use case content appeared with black text even when dark mode was enabled, making it difficult to read. The page background was changing correctly, but the rendered notebook/content text was not inheriting the dark mode text colour.

## Testing

- Checked the use case detail page in dark mode.
- Confirmed the main content text is now visible on the dark background.
- Checked that light mode still displays correctly.


next_webapp/.env
suba-t24, line 1: Please remove this .env file

suba-t24 (changes requested):
The UI/React code changes are fantastic, but you must remove the .env file from the branch before merging this PR

abdullahsher445:
i have removed .env file

suba-t24 (approved):
The code changes are clean, address annoying UI/UX bugs, and use standard React patterns

PR #2036, opened 2026-07-26
feat: add Contributors section to About Us page

Muhib1-lewis:
Adds a Contributors section to the About Us page showing past 
students by year, trimester, team and role.

- Handles varying trimester counts and team names per year (not fixed)
- Fixed a dark mode border styling bug
- Currently using real historical data as a placeholder until the 
  backend API is ready


suba-t24 (dismissed):
The frontend implementation is clean, responsive, and well-structured.
Note: plan a follow-up task to connect this static UI to the new /api/contributors endpoints once they are ready.

suba-t24 (changes requested):
Since the Admin dashboard Contributor update section and Contributor section in About us page involves same Data model. I suggest both section have a common data model instead of creating 2 different ones. Please check with Sher and agree on a common data model if possible

suba-t24:
@Muhib1-lewis if you could just fix these, it would be great.

- role="region" with no aria-labelledby - The accordion panel <motion.div> has role="region" but no label. Screen readers won't announce what region this is. The trigger button needs an id and the panel needs aria-labelledby pointing to it.
- No i18n support - All strings are hardcoded English ("Meet the Teams Behind MOP", "Trimester", "Mentored by", etc.). The rest of the About page uses useTranslations("about") from next-intl. This breaks localisation consistency.

suba-t24 (approved):
Looks good with clean React code and styling

PR #2042, opened 2026-07-27
feat(mongo): add Mongoose models for all collections

lakindubd:
Adds Mongoose models for all 8 collections in mop_database, plus a cached connection helper. This is the Phase 1 (data model) deliverable for the Postgres→MongoDB migration.

Files added:

Connection helper:

next_webapp/src/lib/dbConnect.ts connects the app to MongoDB and reuses the connection instead of opening a new one every time

Models (one per collection):

next_webapp/src/models/mongoose/UseCase.ts
next_webapp/src/models/mongoose/User.ts
next_webapp/src/models/mongoose/Role.ts
next_webapp/src/models/mongoose/Category.ts
next_webapp/src/models/mongoose/Tag.ts
next_webapp/src/models/mongoose/GalleryImage.ts
next_webapp/src/models/mongoose/Blog.ts
next_webapp/src/models/mongoose/Log.ts

Roles: modeled as a collection + embedded ref, not inlined. We'd earlier considered inlining role onto users, but live data showed roles is a real populated collection that users reference via an embedded { id, legacy_id, role_name } object same pattern as usecases.category. Matched the data rather than forcing the inline design. (So 8 collections, not 7.)

Usecase content → GridFS-ready. The schema declares content_file_id + content_type instead of inline content, to get past MongoDB's 16MB limit (usecase id 33 is 22.5MB). The actual GridFS data migration is Phase 2 work see note below.

Timestamps mapped to existing created_at/updated_at where present; omitted for Tag/Log which genuinely lack the paired fields.

Only genuinely-required fields marked required, to avoid rejecting valid existing documents that have nulls.

Not touched: the legacy src/mongodb/ Mongoose files and src/models/Category.ts (Postgres interface) both left for Phase 4 cleanup / separate concern.

### DIFF
--- a/next_webapp/src/lib/dbConnect.ts
+++ b/next_webapp/src/lib/dbConnect.ts
@@ -1,0 +1,60 @@
+import mongoose, { type Connection } from "mongoose";
+
+const MONGODB_URI = process.env.MONGODB_URI;
+
+interface MongooseCache {
+  conn: Connection | null;
+  promise: Promise<Connection> | null;
+}
+
+declare global {
+  // eslint-disable-next-line no-var
+  var __mongooseCache: MongooseCache | undefined;
+}
+
+const cached: MongooseCache = global.__mongooseCache ?? {
+  conn: null,
+  promise: null,
+};
+
+if (!global.__mongooseCache) {
+  global.__mongooseCache = cached;
+}
+
+/**
+ * Cached Mongoose connection helper. Reuses a single connection pool across
+ * hot reloads (dev) and warm serverless invocations instead of opening a new
+ * one per call/import.
+ */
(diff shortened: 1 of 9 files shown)


suba-t24 (approved):
The PR looks solid and I don't see any issues or concerns, so looks good for me!

HarshanaDeakinUni (approved):
looks ok to me

champa1n73 (commented):
# User.ts

- Should add **select: false** property to the password attribute or use `toJson`/`toObject` that omits password.
- Should add **lowercase: true** property to the email attribute.
- Should add an email pattern validation using **match** property in email attribute.

champa1n73 (commented):
hgf

PR #2044, opened 2026-07-29
DS: Use Case Update UC00027

maanak-16:
## What this PR does

This fixes up `UC00027_Spot_Parking_Problems.ipynb` so it matches the new Use Case Notebook template and addresses everything Ved flagged in the quality review.

## Fixes from Ved's review

- **Map tiles were broken** — it was using `Stamen Toner`, which got deprecated and now needs a Stadia Maps API key to work. Swapped it for `CartoDB positron` instead, which just works out of the box.
- **Dependency folder was hardcoded and generic** — renamed `interactive_dependencies` to `UC00027_dependencies` so it's actually tied to this use case, updated the paths in `plot_map()` and `show_interaction()` to match.
- **No evaluation metrics** — the model was being trained and used to predict without ever checking if it was any good. `lm_model()` now prints R² and MAE so you can actually see how well it's fitting.
- **Missing sections from the template** — added Contents, Background, and a proper Data Sets section with working links, since these were required but missing.

## Other things I cleaned up while I was in there

- Updated the title so it actually reflects what the notebook does.
- All the imports were scattered inside random functions — pulled them all into one cell at the top.
- The docstring for `API_Unlimited()` was sitting after the function body instead of being the actual docstring, so it wasn't doing anything. Fixed that.
- Standardised spelling to AU/UK English (analyse, visualise, normalise, etc. — was a mix before).
- The interactivity section had a whole outdated tangent about setting up Binder and `nbinteract` with a GitHub account etc. None of that's needed — it's just standard `ipywidgets` running in the notebook, so I stripped that out and replaced it with a one-line explanation.
- The dataset links (all 4 of them) were pointing to the City of Melbourne's old open data portal and just 404'd. Updated them to the current URLs.
- The API fetch function was loading the whole response into memory as a string before parsing it, which caused a MemoryError on my machine. Changed it to stream straight into pandas instead.
- Added the actual `clue_blocks.geojson` and `merged_data.csv` files into `UC00027_dependencies` since the notebook depends on them and they weren't in the repo.

## Testing

Ran the whole notebook top to bottom (Restart Kernel + Run All) in VS Code. Confirmed:
- All three CLUE datasets pull in fine via the API
- Cleaning, merging, and model training all run without errors
- R²/MAE actually print after training
- The interactive widgets work — year range, checkboxes, impact slider, "Start Analysing" button
- Bar chart, results table, and the map all show up correctly once you run the demo

## Reviewers

@VedS8  @molliefernandez-mentor


molliefernandez-mentor (changes requested):
Hi Maanak, I've got a bit of feedback, the first use case that comes through will take a bit of back and forth but hopefully it will be smoother next time!

- the CSS needs to be kept at the top of the use case as per the template. 
- Please update based on the template and the CSS, you are working on an old version. 
- The contents needs to be after the datasets heading and from scenario - datasets should not be included in the contents page.
- Please delete the key takeaways, this should be reflected already i the 'what this use case will teach you"
- Any dependencies for the file should be kept in the [DEPENDENCIES ](https://github.com/Chameleon-company/MOP-Code/tree/master/datascience/usecases/DEPENDENCIES) folder
- CSS should be applied to all of the headings, and bodies of text. 
- all heading under the contents page should be numbered as per the contents page.
- Please use the colours in the template for the visualisations. 
- Every heading should have an explanation under it explaining what you are doing
- There is no conclusion at the end. 


Start with this and re-request my review once you have implemented it.

maanak-16:
### Summary of changes

- Updated the notebook to fully match the latest MOP template (CSS, sections, headings, Contents, numbering, colour palette, and Conclusion).

- Removed the Key Takeaways section and integrated it into What this usecase will teach you.

- Moved dependencies to DEPENDENCIES/UC00027_Spot_Parking_Problems and updated all paths.

- Added and explained the regression evaluation metrics (R² and MAE).

- Confirmed the notebook uses CartoDB Positron tiles.

- Fixed the map tooltip (area_name instead of the invalid clue_area field).

- Removed an unused column rename in the interactivity function.

- Kept Linear Regression since the target (parking_spaces) is continuous, and clipped predictions at 0 to avoid negative values.

_Note: my branch is currently behind master due to the recent repo cleanup - happy to merge that in if needed, but wanted to flag it rather than risk it locally given the size of the change._

molliefernandez-mentor (changes requested):
Hi, here are a few more feedback points :)

- Usecase-header-bar format needs to be applied to the title of the use case.
- This needs to be removed

Image

- in section 2.4 there is a dataset created and put into the Github and then extracted from the Github, it there a better way to create this dataset without creating an additional dependency, I want to limit the need to reach from the github. Also it definitely needs a better name than 'merged data' if it is necessary, but I suspect there is a better way to merge the data. . 
- In the scenario can you please include the full name of CLUE (Census of Land Use and Employment (CLUE)). 
- Can you please update the name of the second dataset, the link says it is called: "Jobs per ANZSIC for blocks" not Employment by city block.
- Same with the third data set, it is being listed as named "Off-street car parks with capacity and type"
- same with dataset 4 should be "Blocks for Census of Land Use and Employment (CLUE)"
- Is there a reason there is a need for the file in the dependency folder at all, given that they are all Melbourne open datasets? Couldn't they just all be called using an API?
- The final file needs to be in the MOP-Code/usecases/FINALISED folder not in update needed.

maanak-16:
Hi Mollie, thanks for the second round of feedback! I've addressed all the requested changes.

**Updates made**

- Applied the usecase-header-bar to the title and removed the leftover template placeholder text.

- Removed all dependency files by reusing the in-memory merged dataset and fetching CLUE boundaries directly from the Melbourne Open Data API, so this use case no longer requires a DEPENDENCIES folder.

- Updated the Scenario to include the full name of Census of Land Use and Employment (CLUE).

- Corrected all dataset names to match the Melbourne Open Data portal.

- Moved the completed notebook to usecases/FINALISED/UC00027_Spot_Parking_Problems/.

molliefernandez-mentor (changes requested):
Thanks Maanak, Looks really good, just a minor naming change and the Index update. 

- The name of the use case is obviously "Spot Parking Problems" and the title should be the same as the name of the file but the title has the extra bit in it: "Spot Parking Problems: Identifying Off-Street Parking Shortfalls in the City of Melbourne" I wonder whether we could change the name so that it is not quite as long as the full title but still includes some of the context and rename the file and title with a new name. I think that might be good, it will certainly help as there are a lot of other use cases that deal with parking problems, I think giving the name some more context will set it apart. 
- Can you please update the Use case Index file as well!

@maanak-16 Once you have finished making changes please **re-request my review**

maanak-16:
**Addressed this round of feedback:**

- Updated the use case title to "Spot Parking Problems: Melbourne's Off-Street Shortfalls" for better consistency with the filename while maintaining clarity and uniqueness.

- Updated the Use Case Index by moving the entry from "Update Needed" to "T2 2026" and pointing it to the new usecases/FINALISED/ location.

molliefernandez-mentor (changes requested):
Sorry, I don't think I was clear with the request to change the name of the use case. I want the name of the file and the title of the use case to be exactly the same and there shouldn't be a : in there.
The name of the file needs to be "Off_Street_Parking_Gaps.ipynb" 
The title in the use case needs to be Off-Street Parking Gaps

Looks like you have put the .ipynb file into a folder that doesn't exist yet, can you please rebase your branch to pull the latest changes (there have been a lot). and move your file into the FINALISED folder (it should not be in it's own folder), just the file by itself is fine!

I also cannot see the changes in the Use case Index file. once you rebase your branch you will see how you need to update it. There is a new table at the top of the file. 

Please remove the bullet points next to the sub-sections in the contents list. 
Image

The conclusion and sections within the conclusion need to link back to the scenario at the start of the use case and talk about how the analysis has achieved the problem stated in the scenario.

maanak-16:
**Addressed this round of feedback:**

- Renamed the file to Off_Street_Parking_Gaps.ipynb and updated the title to "Off-Street Parking Gaps".
- Rebased onto latest master, moved the notebook directly into usecases/FINALISED
- Updated Use Case Index to the new Finalised table format
- Removed the bullet points next to the Contents sub-sections 
- Linked the Conclusion and each of its sub-sections back to the Scenario's stated problem, explaining how the analysis addresses it

_Note: I noticed UC00010_2023_Melbourne_Liveability_Assessment in the Finalised table keeps its UC-number prefix in the filename — let me know if you'd like this one to follow that pattern too, happy to adjust either way._

s224251967-hub (dismissed):
Hi, a few feedback comments as below:

**- Verify Date Created:** Please verify the original creation date. “Trimester 2, 2026” appears to describe this update rather than the original use case.
**- Clarify CLUE and datasets:** “CLUE” refers both to the Census of Land Use and Employment framework and to the Blocks for Census of Land Use and Employment dataset. This is confusing. Please use the terms consistently and briefly describe each dataset’s role.
**- Revise Section 1 explanation:** Please focus this explanation on what the code does, rather than on matching MOP template colours.
**- Name the first dataset:** Please specify that the first dataset fetched is the Residential Dwellings Dataset. “The first of our three datasets” is unclear because four datasets are listed.
**- Explain the 2021 exclusion:** Please explain why 2021 is excluded from Residential Dwellings data. This removes over 10,000 records and is not explicitly applied to the other datasets.
**- Handle zero Impact index:** Please guard against a zero denominator when normalising Impact_index. If no positive shortfall exists, max(impact[impact >= 0]) can be zero and the calculation may fail.
**- Avoid hard-coded 2020:** Please avoid referring to 2020 as fixed in headings, explanations, and comments
**- Merge Demo use case:** Please remove the “Demo use case” section and move it‘s Contents entry to "Section 5.2 Interactivity", like requested in the previous review.

Optional additional comment:
**- Confirm analysis period:** It's better to confirm the common available year range across the three analytical datasets and state the analysis period clearly in the notebook.

maanak-16:
**Addressed this round of feedback from Xiaolin:**

- Clarified CLUE terminology and added a short role description for each dataset.
- Reworked Section 1 to focus on what the code does rather than template styling.
- Explicitly named the first dataset as Residential Dwellings.
- Investigated the 2021 exclusion and removed the filter after finding no data-quality justification.
- Fixed a divide-by-zero bug in Get_predict_result().
- Removed all hard-coded 2020 references and made the year slider dynamic.
- Merged Demo use case into Section 5.2 as previously requested.
- Added the common analysis year range across all three datasets.
- Updated the Date Created

molliefernandez-mentor (changes requested):
HI @maanak-16, 

- With the date created it still needs to be Trimester and year, so just update it to be T2 2024.
- And yes with the file name it should still have the UC00027 code at the front!
- Same with in the use case index including the code

maanak-16:
Hi Mollie,

Renamed the file to include the UC00027 code, and updated Date Created Format."

molliefernandez-mentor (approved):
Thanks @maanak-16 looks good from me, wait for @VedS8's final review then you can merge. 

I would also recommend rebasing your branch before merging as there have been quite a few changes since the PR was made - there are also some conflicts with the use case-index, and rebasing should fix this.

PR #2045, opened 2026-07-30
feat(contributors): Contributors API model + CRUD endpoints

lakindubd:
A MongoDB model and a full CRUD API for contributors.

- src/models/mongoose/Contributor.ts the data model. Flat structure (one document per person), with the agreed team/role taxonomy enforced as enums. Exports TEAMS (4), ROLES (10), LEVELS, CONTRIBUTOR_TYPES so the admin form can import the same lists instead of hardcoding them.

- src/app/api/contributors/route.ts: GET (list all, public) and POST (create, admin-only).

- src/app/api/contributors/[id]/route.ts: GET, PUT, DELETE for a single contributor.

### DIFF
--- a/next_webapp/src/app/api/contributors/[id]/route.ts
+++ b/next_webapp/src/app/api/contributors/[id]/route.ts
@@ -59,8 +59,11 @@ export async function PUT(
     let body;
     try {
       body = await request.json();
-    } catch {
-      return errorResponse("Invalid JSON in request body.", 400, "INVALID_JSON", request);
+    } catch (error) {
+      if (error instanceof SyntaxError) {
+        return errorResponse("Invalid JSON in request body.", 400, "INVALID_JSON", request);
+      }
+      throw error;
     }
 
     await dbConnect();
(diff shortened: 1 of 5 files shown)


HarshanaDeakinUni (approved):
the logic is sound and middlware paths are updated. well documented too so all good there, lgtm.

champa1n73 (changes requested):
# contributors

- Method `toContributorDTO` is duplicated in both files `next_webapp\src\app\api\contributors\[id]\route.ts` and `next_webapp\src\app\api\contributors\route.ts`. I suggest extract to a shared util to avoid drift.
- In `next_webapp\src\app\api\contributors\route.ts`, `const body = await request.json();` can throw on malformed JSON. Please catch that `SyntaxError` and return a `400` (`INVALID_JSON`) instead of falling through to the generic `500` error response.
- `errorResponse()` accepts `request?`, but route handlers do not pass the request object.
  - As a result, logs miss request context (`url`, `method`, `ip_address`, `user_agent`).

# middleware.ts

- `verifyJWT' method should validate header metadata. It verifies signature and expiry, but does not confirm `alg === "HS256"` or `typ === "JWT"`.

champa1n73 (changes requested):
- In `POST`/`PUT` of `next_webapp\src\app\api\contributors\[id]\route.ts` and `next_webapp\src\app\api\contributors\route.ts`, the JSON parse block is too broad. Change `catch { ... }` to `catch (err)` and only return `400 INVALID_JSON` when `err instanceof SyntaxError`; all other errors should fall through to the outer handler.

champa1n73 (approved):
Everything is working fine.

PR #2049, opened 2026-08-02
Phase 2 – Migrate GET /api/search from Supabase to MongoDB

suba-t24:
This PR:

- Rewrites the search API route to use dbConnect() and the UseCase Mongoose model. 
- Fixes a data-leak where the old route returned raw notebook content in search results (select('*') → explicit field whitelist). 
- Adds sortBy/sortOrder support. 
- Rewrites the test suite to mock Mongoose instead of Supabase (13 tests passing). 
- Also fixes broken filter behaviour in the search page UI (locale-aware routing, draft state for all filters, single Apply Filters button).


next_webapp/src/components/SearchFilter.tsx
bettycheng14, line 1: The updated `Category` Mongoose schema (src/models/mongoose/Category.ts) has no `slug` field. 
I think the Category interface here is outdated.
Suggestions to update on this file:
1. Update interface Category (line 6-13)
2. Use `category.legacy_id` or `category._id` instead of `category.slug` as the selection state identifier. (line 130)

bettycheng14 (changes requested):
All 5 objectives are met. Nice catch on the security vulnerability, and test coverage looks solid. The UI filter refactor is clean overall, just needs the `category.slug` fix noted above before it's fully working.

next_webapp/src/components/SearchFilter.tsx
suba-t24: Fixed and pushed the changes

bettycheng14 (approved):

HarshanaDeakinUni (approved):
the recomendations betty suggested were applied and fixed so nothing else to add from me, lgtm now,

PR #2051, opened 2026-08-02
converted about page to Server component and reduced client bundle size

Muhib1-lewis:
Based on suba's feedback

1. Fixed the real issue: about/page.tsx still had "use client" left 
  over, so the Contributors JSON was being bundled into client-side 
  JS unnecessarily
2.  Made page.tsx a proper Server Component and pass the data down as 
  a prop instead


suba-t24 (approved):
Good fix — removing "use client" from page.tsx is the right approach. One note: ContributorsSection itself still has "use client", so the JSON data crosses the server→client boundary as a serialized prop. The import is no longer in the client module graph, which is the meaningful win. Also, the as unknown as ContributorRecord[] cast on line 9 is worth revisiting if the type doesn't match the JSON shape — a zod schema or fixing the type definition would be cleaner long-term.

PR #2052, opened 2026-08-02
WD: Fix broken use-case search paths and remove duplicate UC00124 folder

hamizhassan16:
- Fixed two Next.js API routes (`next_webapp/src/app/api/route.ts` and `next_webapp/src/app/api/search-use-cases/route.ts`) that still pointed to the old `datascience/usecases/READY TO PUBLISH` path. That folder no longer exists after the recent restructure that moved `datascience/` contents to the repo root, so use-case search was silently returning no results.
- Removed a duplicate copy of `UC00124_Environmental_Resource_Consumption_and_Service_Demand_in_Melb`, which existed identically under both `Urban_Planning_and_Development/2024/T3` and `Environmental_Sustainability/2024/T3`. Verified all three files (`.html`, `.json`, `.ipynb`) were byte-for-byte identical before removing the `Urban_Planning_and_Development` copy.


suba-t24 (dismissed):
Thanks for catching this and for the duplicate removal - that part is correct.
However the path fix in both routes resolves to repo-root/usecases/READY TO PUBLISH instead of repo-root/datascience/usecases/READY TO PUBLISH. The notebooks are still under datascience/. The correct path for both files should be:
               path.join(process.cwd(), "..", "datascience", "usecases", "READY TO PUBLISH")

search-use-cases/route.ts was already correct on master (it had "..", "datascience") - this PR accidentally removes "datascience" from it. Please restore both to "..", "datascience", "usecases", "READY TO PUBLISH".

hamizhassan16:
> Thanks for catching this and for the duplicate removal - that part is correct. However the path fix in both routes resolves to repo-root/usecases/READY TO PUBLISH instead of repo-root/datascience/usecases/READY TO PUBLISH. The notebooks are still under datascience/. The correct path for both files should be: path.join(process.cwd(), "..", "datascience", "usecases", "READY TO PUBLISH")
> 
> search-use-cases/route.ts was already correct on master (it had "..", "datascience") - this PR accidentally removes "datascience" from it. Please restore both to "..", "datascience", "usecases", "READY TO PUBLISH".

I think you are mistaken , you need to pull changes to have latest master branch in which the data science folder was changed by me in previous PR, all folders inside datascience folder have been moved to repo root and the folder doesn't exist. these changes are correct in my opinion, kindly check the latest master branch and see, still if any discrepancies are there , kindly let me know :)
Screenshot 2026-08-03 at 12 54 45 PM

molliefernandez-mentor (changes requested):
Looks great!

molliefernandez-mentor (approved):
Looks Great

PR #2053, opened 2026-08-02
Migrate use cases to MongoDB with GridFS notebook storage

lakindubd:
What this does

Moves the entire **use cases** feature off Supabase/Postgres and onto MongoDB, and changes how notebook files are stored. Reads, creates, edits, and deletes for use cases now all go through MongoDB. Admin-only actions (create/edit/delete) are now properly locked behind login; viewing stays public.

Why

problems with the old setup:

 **Notebooks were stored inline in a database column.** MongoDB has a hard 16 MB limit per record, so any large notebook silently failed to save (this is why the 22.5 MB "Night Time Safety Index" use case never migrated). Storing notebooks in **GridFS** (MongoDB's system for large files) removes that limitation.

How notebooks are stored now

- When a use case is created or edited, the notebook file is saved into GridFS and the use case record just points to it.
- When a use case page loads, the notebook is streamed back out of GridFS and rendered same as before from the user's point of view.
- **Uploads over 60 MB are rejected** (there was previously no size limit at all).
- When a use case is deleted, or its notebook is replaced, the old file is cleaned up so storage doesn't leak.
- Content is stored using a **"mint new" approach**: editing a notebook creates a fresh file and only deletes the old one after the save succeeds so a use case never ends up pointing at a missing file if something fails halfway.

Files added

- `src/lib/gridfs.ts` helper that hands out a GridFS connection for notebook storage.
- `src/app/api/usecases/[id]/content/route.ts` new endpoint that streams a use case's notebook out of GridFS.
- `src/app/api/usecases/_shared.js`  shared logic (validation, size cap, file upload/cleanup, tag/category resolution) used by both create and edit, so the two can't drift apart.

Files changed

- `src/app/api/usecases/route.js`  list + create rewritten for MongoDB/GridFS.
- `src/app/api/usecases/[id]/route.js`  get + edit + delete rewritten for MongoDB/GridFS.
- `src/app/[locale]/usecases/[id]/page.tsx`  use case page now fetches notebook content from the new streaming endpoint and decides how to render based on a stored content type, instead of guessing.
- `src/middleware.ts`  `/api/usecases` write actions now require an admin login; reads stay public. Matches the same pattern the contributors feature already uses.
- `src/models/mongoose/*.ts` (all models)  small type-only fix to the model exports (no behaviour change), needed so the routes type-check cleanly.

Files removed

Deleted dead, unused legacy MongoDB files that were left over from an old prototype. They were harmless while `MONGODB_URI` was unset, but this PR sets that variable live which would have "woken up" this old code. Confirmed nothing in the app imports them before deleting:

- `lib/mongodb.js`  an old duplicate database-connection helper (the real one is `src/lib/dbConnect.ts`, untouched).
- `models/User.js`  an old unused User model (the real one is `src/models/mongoose/User.ts`, untouched).
- `src/mongodb/`  a folder of early prototype files, none referenced anywhere.

What's intentionally NOT in this PR (follow-ups)

- Loading the Data Science team's new `finalized/` notebook set this PR builds the pipe, the actual content load happens once that's delivered.
- Categories aren't migrated to MongoDB yet, so a new use case's category is stored as a reference to be resolved later.
- Deleting the now-unused `/api/usecases/[id]/tags` route and converting the two route files to TypeScript, minor cleanup.


suba-t24 (approved):
The mint-new GridFS pattern and doc-first DELETE ordering are both correct, auth checks are properly enforced on all write routes, and the dead legacy files are safely removed. The JWT algorithm-confusion fix is a nice security bonus.

Minor follow-ups (non-blocking): convert _shared.js to .ts, and the content endpoint Cache-Control can be immutable since edits always mint a new file ID.

Good work on a clean migration!

PR #2054, opened 2026-08-03
DS: Cleanup outdated Playground folders

adhikari8848:
Removes student folders from previous trimesters that are no longer active, so the Playground reflects the current T2 2026 team.

42 folders were removed: 37 from the Playground/ directory and 5 from the Playground/ARCHIVE/ directory. No folders were deleted simply based on their name—each notebook was first compared against the corpus in the usecases/ folder and was only deleted in cases where a verified copy still existed, with the copy matching both in file content and in source cells, or with at least 85% of the code cells being present in the published notebook corresponding to the same use case ID.

42 folders are deliberately left in place. They belong to students no longer on the team but hold the only copy of 60 notebooks that exist nowhere else, including 10 folders still under ARCHIVE/.

All 12 current team member folders are untouched.

Nothing outside Playground/ is affected: 195 files deleted, none added or modified.

I have tracked these changes using the Excel sheet:
[MOP_Playground_Cleanup_Tracker.xlsx](https://github.com/user-attachments/files/30641805/MOP_Playground_Cleanup_Tracker.xlsx)


molliefernandez-mentor (approved):
Looks good! @adhikari8848 Please merge the PR and delete the branch

PR #2055, opened 2026-08-03
DS: Usecase Template Update

AtishayDeakin:
Removed numbering styles for section headings and updated CSS rules for use case templates.


molliefernandez-mentor (approved):
Thanks :)

PR #2059, opened 2026-08-03
WD: Migrate statistics route to MongoDB

qweasdzxc03:
- Count, dataset, tags, trimester routes deleted 
- Category and total count routes updated


bettycheng14 (changes requested):
Great work! The aggregation pipeline is a huge performance upgrade and removing the dead routes was a great call. Just three quick tweaks needed:

1. Root Dependency: It looks like the `mongodb` package was accidentally added to the workspace root `package.json`. Please revert those root changes and merge the latest master into your branch so it doesn't block the PR with merge conflicts.

2. API Contract: The frontend dashboard expects `{ category: "..." }` to filter in `src/app/[locale]/admin/dashboard/page.tsx`. Your aggregation returns `category_name`. Please map it back to `category` in `$project` so the dashboard doesn't break.

3. Missing Fallback: In MongoDB, use cases without a category are grouped under _id: null. Please use $ifNull in your $group stage to safely fall back to "Uncategorized" like the old codebase do.

qweasdzxc03:
Updated! Please have a look.

hamizhassan16 (changes requested):
Nice work — the migration to MongoDB for by-category and total-count looks solid, and I verified the 4 removed routes (count, dataset, tags, trimester) aren't used anywhere in the frontend, so that cleanup is safe. Good catch fixing the category field name to match what the dashboard expects.
One small thing before merging: looks like npm install mongodb accidentally ran from the repo root instead of next_webapp/ — mongoose is already installed there, so it wasn't needed. The root package.json was reverted, but package-lock.json still has leftover mongodb entries. Could you check that file too? Otherwise, good to go

suba-t24 (dismissed):
The migration is correct, the aggregation matches the schema, the deleted routes are genuinely dead, and the cleanup is needed.

- One thing to add: export const runtime = 'nodejs' at the top of both updated route files, to match the pattern used in the rest of the Mongoose routes and prevent edge runtime errors
- Also note the files use Windows line endings (\r\n) — worth normalising to \n if your editor supports it.

bettycheng14 (approved):
Checked through all the feedback and requested changes. Everything's been addressed nicely.

- ✅ Root package.json and package-lock.json removed
- ✅ Aggregation correctly maps to category with $ifNull fallback to "Uncategorized"
- ✅ `export const runtime = 'nodejs'` added to both route files
- ✅ Line endings normalized to `\n`
- ✅ Dead routes (count, dataset, tags, trimester) cleanly removed

This pr is ready to merge!

hamizhassan16 (approved):
Mongodb issue I flagged is fully resolved. All good from my side too ,approved

PR #2061, opened 2026-08-03
DS: Usecase Update UC00163

aidanuni:
## Summary
Update of UC00163 (Parking Availability Melbourne CBD, originally authored by Stefan Cucos T1 2025) for deployment. Moved to usecases/FINALISED/. 

## Template alignment
- Converted to the updated use case template. New CSS classes, metadata structure and unnumbered front/end matter
- Plot colour convention applied
- Removed emoji, standardised AU spelling

## Rebuild
The original notebook's ML section scored 50% on a binary target because it lacked a temporal signal. The analysis was reframed on the archived 2019 CoM parking sensor dataset (42.7M events, filtered to March 2019 with 8 core CBD areas in `usecases/DEPENDENCIES/`):
- Availability measured as vacant bay-minutes via the VehiclePresent column
- Duration-weighted Random Forest: 0.703 accuracy vs empirical lookup 0.685 and majority baseline 0.594
- `predict_availability()` function answers the original user story
- Live merge fixed on the documented kerbside/bay-id key (2.2% match rate attributed to the ongoing CoM parking system upgrade)

## Housekeeping
- Notebook re-executed end-to-end before commit
- `Use_Case_Index.md` updated with a Finalised section and UC00163 row


molliefernandez-mentor (changes requested):
Hey @aidanuni, please make the changes we discussed in our 1-on-1. Then re-request my review.

aidanuni:
Thanks for the review Mollie, this addresses the points you raised.

Template and styling

Sections and sub-sections are now numbered with id anchors and the Contents list links to them
All text wrapped in usecase-body / usecase-list divs with heading cells separated from body cells
Header bar and plain-English title added conclusion sub-headings converted to template classes

Readability

Output explanations rewritten for a non-technical reader rather than describing data shape and structure

Reliability

Added retry logic and optional API key support via .env following the UC00157 pattern.

molliefernandez-mentor (changes requested):
Hi @aidanuni I currently can't open the use case (see error below) so I can't do a full review but based on the version that opens in Git here si some feedback

Image


- Please update your CSS - it is an old version
- Your file name should have underscores between each word and it should be the same as your usecase title (include "in the")
- remove your name from the authored by seciton
- remove the 'historical (archive)' heading (and the 'live' dataset heading), just list the dataset like the others (it makes it seem like it is no longer in use)
- the 'note' section should be included with the sentence above about the datasets. 
- In the conclusion can you please add a bit more in the top section about what the result was that you actually achieved? What was the outcome of the analysis?

aidanuni:
Thanks Mollie, all six addressed:

CSS replaced with the current usecase_TEMPLATE version
Renamed to UC00163_Parking_Availability_in_the_Melbourne_CBD.ipynb, matching the title
Removed my name from Authored by (still shown as the updater in the index row, matching the other Finalised entries)
Datasets now listed flat with no live/historical sub-headings and the note folded into the sentence above the list
Conclusion opens with the actual outcome: the daily availability pattern, the model result against baselines and the working prediction function
Contents converted to the template's numbered-list style while I was in there

molliefernandez-mentor (changes requested):
Thanks @aidanuni, looks great. Just remove the dot points from the contents list. I am also running the code and it seems to be taking a very long time, I stopped it at the 2 hour mark but it needs to be able to run in a reasonable amount of time. Box 14 was the one it was getting stuck on, where the data was being loaded from the DEPENDENCY folder.
Also can you please put the file in the dependency folder into a folder with the name of the use case.

aidanuni:
Thanks Mollie. All three sorted,

Bullet points removed from the Contents so sub-items are now indented links

The extract has moved to usecases/DEPENDENCIES/UC00163_Parking_Availability_in_the_Melbourne_CBD/

On the load time, the load cell had a fallback that searched the whole usecases/ tree recursively if its explicit paths missed. On my laptop the first path always matched so the search never ran, but on any setup where it missed it would go for hours instead of failing. I've removed the fallback entirely so it now checks four paths and raises straight away with the expected location if none match. The read itself is about half a second for the 508k rows, so the cell should be practically instant now.

molliefernandez-mentor (approved):
Thanks Aiden, 

I can't run the code until your dataset is actually in the dependency folder, otherwise there is no data to pull, so can you please merge the PR and then I'll check that the code runs fine. 

Please make sure you rebase your branch before merging, there has been a few changes to the repo.

PR #2063, opened 2026-08-04
WD: fix(usecases): migrate recent route to Mongo, remove dead tags route , fix admin notebook removal

lakindubd:
While checking the use cases migration state on master, found a few things under `/api/usecases/**` that got left behind when the resource moved to MongoDB two routes were still on Supabase and are now broken for any use case created after the switch, plus a bug in the admin edit page. This PR cleans all three up.

**1. `/api/usecases/[id]/tags` was broken and redundant**
This route still queried the old Supabase tags table and ran `Number(id)` on the use case id which fails for every MongoDB id (they're 24-character strings, not numbers), so it returned a 400 on any current use case. It was also unnecessary: the main `GET /api/usecases/[id]` already returns the tags directly. Removed the route and pointed the admin edit page at the tags it was already receiving.

**2. `/api/usecases/recent` (the home page "recent use cases") was still on Supabase**
The home page's recent-use-cases section was reading from the old database, so newly created use cases didn't show up correctly, and the "View Details" links were pointing at ids that no longer match. Migrated it to MongoDB, matching how the other use case routes work. No frontend change needed it returns the same fields the home page already reads.

**3. Admin "Remove notebook" button didn't actually remove the notebook**
Two problems stacked on top of each other:
- Clicking "Remove notebook" sent an empty signal the backend read as "no change," so the notebook was never actually cleared.
- The button didn't even appear for existing notebooks in the first place it was checking the wrong field to detect whether a notebook was present (it looked for the notebook content on the record, but that now lives in GridFS, so the check was always false).

Fixed both: the button now shows correctly for existing notebooks, and removing one properly clears it and deletes the stored notebook file.

Files changed

- **Deleted** `api/usecases/[id]/tags/route.ts` broken + redundant (plus its now-orphaned test, which would have failed the whole test run).
- **Rewritten** `api/usecases/recent/route.ts` migrated to MongoDB.
- **Edited** `admin/use-cases/edit/[id]/page.tsx` reads tags from the main use case response, and the notebook-removal fix.


hamizhassan16 (approved):
Nice, clean migration. Also appreciated catching the uc.content → uc.content_file_id bug and the null-vs-undefined fix for clearing notebooks, both are correct against the backend's actual semantics. Approved from my side.

suba-t24 (approved):
All three fixes are correct, well-reasoned, and safe. The deleted routes were genuinely broken, the Mongo migration of recent is clean, and the null/undefined/content_file_id distinction in the edit page correctly wires up to the existing backend protocol.

PR #2065, opened 2026-08-04
DS: Usecase Update UC00088

adhikari8848:
Updates UC00088 to the new template and adds the analysis:  
Distance from each pedestrian sensor to its nearest bus stop,
stops within a 400 m walk, and a correlation against foot traffic.

Previously, the two datasets were only plotted on the same map.

Also fixes — the API column names have changed since 2024.

Moved to usecases/FINALISED. Original use case by Shilpa Sharma.


molliefernandez-mentor (changes requested):
Thanks Sumit, Here is some feedback 

- please remove the "UC00088" from the title of the use case
- delete this sentence under datasets "Both datasets come from the City of Melbourne Open Data Platform and are downloaded live through the Explore API, so the notebook always uses the current version"
- move this section to just below the data sets heading and remove the "Worth knowing about the pedestrian data"
- remove this section

Image

- Change 1.2 to: "Using the API to call the data from the Melbourne Open Data Site"
- Combine the "What you can take away from this" section into the "Observations for further opportunities" section
- in the "what we learnt from this analysis" section please use the unnumbered list formatting.
- Combine this section below into the "what we learnt from this analysis section:

Image

- Please update the usecase index as shown in the Master branch (at the top of the file)

@adhikari8848 Once you have finished making changes please **re-request my review**

adhikari8848:
@molliefernandez-mentor 
Hi Mollie, I have implemented the requested changes. Please review them. Thank you.

molliefernandez-mentor (changes requested):
Replace the highlighted sentence with the below text
Image
"The analysis found that Melbourne’s busiest pedestrian locations are generally well served by nearby bus stops. However, this does not necessarily mean the stops were located based on pedestrian activity, and six high-pedestrian locations were identified as exceptions where bus access is limited."


Otherwise it's looking really good!

adhikari8848:
@molliefernandez-mentor 
Hi Mollie, I have updated the paragraph; please review them. Thank you.

molliefernandez-mentor (approved):
Thanks @adhikari8848 looks good, happy for you to merge once you get a review from @VedS8

VedS8 (approved):
Great work, approving for merge!

PR #2067, opened 2026-08-05
WD: Global CSS & Dark Mode Stability

sahansaa:
## Summary
Establishes a unified global CSS foundation and fixes dark mode inconsistencies across the About, Login, and Privacy pages.

## Changes
- **tailwind.config.ts**: No change needed — `darkMode: 'class'` was already configured correctly.
- **layout.tsx**: Removed the conflicting `Inter` font import/usage. Poppins (already set in globals.css and tailwind.config.ts) is now the single active font project-wide.
- **globals.css**: Removed the body background rule referencing undefined `--foreground-rgb` / `--background-*-rgb` variables, which caused a translucent backdrop. Replaced with the project's actual `--color-dark` / `--color-light` variables, plus a new `.dark body` rule.

## Bug found during testing
Dark mode toggle logic previously lived only in the Privacy page's local state + localStorage effect. About and Login had `dark:` classes defined but no way to activate them on a fresh page load, dark mode only appeared to work on those pages after client-side navigation from Privacy. Fixed by adding a blocking inline script in `layout.tsx`'s `<head>` that reads the same localStorage `"theme"` key on every page load, before paint.

## Testing
Verified locally:
- About, Login, and Privacy pages render a solid background (no translucent artifacts) in light mode
- Poppins is the only font active across all three pages
- Dark mode toggle on Privacy page still works and now persists correctly on a hard refresh of About/Login
- No console errors on any of the three pages

## Acceptance criteria
- [x] `tailwind.config.ts` contains a defined dark mode strategy
- [x] Only one primary font family is imported and active across the entire application
- [x] The body background renders a solid color without referencing missing RGB variables


suba-t24 (changes requested):
The layout.tsx and globals.css changes are correct — removing the Inter font conflict, fixing the undefined --foreground-rgb variables causing the translucent background, and the blocking inline script is the right approach for preventing dark mode flash on page load (uses the correct "theme" key consistently with the rest of the app).

However, the branch contains many unrelated files that cannot be merged: .DS_Store, .vs/ Copilot index databases, .vscode/settings.json, yolov8n.pt, UC00221_Urban_Flood_Risk_Prediction.ipynb, stray pr_note.txt files, and others. Please open a clean branch with only the three intended CSS/layout changes and reopen the PR.

PR #2068, opened 2026-08-05
DS: Usecase Update UC00157

AtishayDeakin:
1. Fixed spelling mistakes and grammar.
2. Removed empty cells.
3. Improved markdown consistency.
4. Aligned fully with the latest template.
5. Ran the notebook sequentially.


molliefernandez-mentor (changes requested):
Hi Atishay, thanks for the PR, here is some feedback:

- I feel like the first paragraph of the scenario should be in the background section.
- Make sure your Melbourne datasets take people to the information part of the dataset not API (/information - at the end)
- update the pedestrian counting dataset to be the actual name of the dataset. 
- Please remove the bullet points that are next to your sub-sections in the contents page. 
- This is a template prompt, please remove it
Image

- the numbered list under 3.2 should be unnumbered 
- can you add why the checks run in 3.2 are necessary, like explain why they are being done, rather than just what they are showing. 
- in 3.3 explain why the features are removed? why are they not relevant? why are you adding the new column? and assigning a value of 1?
- I feel like 3.4 needs a map background or something?
- same feedback from 3.2 applies to 4.2
- Throughout the use case I would love to see more reference to what is being analysed and the results and how they relate back to the scenario at the start of the usecase. 
- under section 10 the list should be unnumbered
- your legend on 11.4 doesn't really show all the heatmap colours 
- Please delete the Night time safety index folder from the ready to publish folder
- I also feel like it needs a different name...safety index is a bit broad when it is focused on pedestrians....

AtishayDeakin (commented):
Hey @molliefernandez-mentor, 

I have done the requested changes and changed the name of file to be more specific as well rather than just a broad category.

molliefernandez-mentor:
Make sure you re-request review next time.
image

molliefernandez-mentor (changes requested):
Hi Atishay, just a few more changes, mostly around the written content and having it be more explanative rather then 

- The link to the usecase in the usecase index file doesn't work (even on your branch).
- Please change the name of the use case to "Night Time Pedestrian Safety Analysis" and change all the references to the 'safety index'
- Throughout the use case I want to see more reference to what is being analysed and the results and how they relate back to the scenario at the start of the usecase. 
- All of the overviews and dot points state facts and outcomes but none of them explain why it is important or how it related back to keeping pedestrians and commuters safe at night. 
- There is an error, the code won't run
Image

AtishayDeakin (commented):
Hey @molliefernandez-mentor, done with the changes.

1. About the usecase index link, it will work when the file is merged in master branch because at the time of merge the link I am providing is the one that will be on master branch.
2. Also, about the error, there was a new installation of library, so I installed it when I was running the notebook on my system but do you want me to add it somewhere? Or will the library be automatically installed?

molliefernandez-mentor (changes requested):
@AtishayDeakin the package needs to be installed using pip! install, and obviously you don't want this to run every time the code is run, you need to put it in a separate code block and explain to the user how to install the package and then how to comment it out once it is installed. 

Could the 3.4. Geospatial visualisation of street lights map be turned into a map that is more interactive? It is a bit funny looking. It also needs a legend. 

The last paragraph you have under 9 is exactly what I'm looking for in terms of linking it back to the scenario, can you please include similar linking sentences for Sections 4.2, 4.3, 5.2, 6.3, 6.5, 7.2 and in the conclusion.

The 11.3 graph needs a y axis label

For the 6.5 graph could you possibly use the name of the location rather than the location ID?

AtishayDeakin (commented):
@molliefernandez-mentor, 

I have done the requested changes.

VedS8 (changes requested):
Hey Atishay, great work - just a couple of things!

1. In section 10-11, it joins all the datasets by rounding coordinates into tiny grid squares (~11m each). Since the datasets are so different in size, most squares only end up with data from one dataset, but the other six get filled in as 0 anyway and averaged in. So most final scores are really just one real number dragged down by six fake zeros, not an actual combined score. Kinda breaks the main point of the analysis; needs fixing before merge. Maybe try using bigger grid squares so more datasets actually overlap, or only average across the datasets that actually have data for that spot instead of counting the missing ones as 0. Open to whatever you think works better, though.

2. Better to just remove the print(load_dotenv()) in cell 1 completely; it just spits out False with no explanation and doesn't add anything.

3. Section 11.2, the numbers in the writing don't match what's printed right above it (says 20,573 grids/mean 0.48/max 1.29, code shows 17,427/0.4558/1.1429), just needs updating to match.

4. Plot colours don't really cycle through the template palette across the notebook. Most individual charts (street lights, pedestrian counts, landmarks, bus stops) just stay green, and purple/orange/blue only show up in the one combined map near the end. Other use cases assign each new chart the next colour in the palette instead of defaulting to green throughout, worth doing the same here.

Follows the template nicely otherwise!

AtishayDeakin (commented):
@VedS8,

I have done the requested changes, please have a look at it. Thanks.

VedS8 (changes requested):
Hey Atishay, this is really solid, the merge fix, grid sizing, and light density logic are all spot on now, and the data-coverage caveats in section 11 are handled really well.

Apologies, I overlooked this in my previous PR review: one small formatting thing in the Conclusion - the subheadings "What we achieved in this analysis," "What we learned from this analysis," and "Observations for further opportunities" run straight into their paragraph text with no gap between them. Worth adding a bit of spacing there so each section is visually separated, same as the rest of the notebook.

AtishayDeakin (commented):
@VedS8, 

I have updated the file again.

molliefernandez-mentor (approved):
Thanks @AtishayDeakin looks good from me, wait for @VedS8's final review then you can merge.

I would also recommend rebasing your branch before merging as there have been quite a few changes since the PR was made - there are also some conflicts with the use case-index, and rebasing should fix this.

VedS8 (approved):
Great work, Atishay! Happy to merge.

PR #2074, opened 2026-08-08
WD: resolve duplicate locale in Back to use cases button href

abdullahsher445:
Description
Fixes a routing bug on the Use Case detail page (usecases/[id]/page.tsx) where clicking "Back to use cases" resulted in a 404 Page Not Found error.

🛠️ Root Cause & Fix
Root Cause: The button used href="/en/usecases" with @/i18n-navigation's Link component. Because next-intl automatically attaches the locale prefix (/en) to all internal routes, it generated /en/en/usecases (duplicate locale), causing a 404.
Fix: Updated href="/en/usecases" to href="/usecases" so next-intl correctly resolves the URL to /en/usecases.
🧪 Testing
Navigated to a Use Case detail page (/en/usecases/[id]) and clicked "Back to use cases".
Confirmed it now successfully returns to /en/usecases without any 404 errors.
image
image

### DIFF
--- a/next_webapp/src/app/[locale]/usecases/[id]/page.tsx
+++ b/next_webapp/src/app/[locale]/usecases/[id]/page.tsx
@@ -172,7 +172,7 @@ const UseCasePage: React.FC = () => {
           )}
 
           <Link
-            href="/en/usecases"
+            href="/usecases"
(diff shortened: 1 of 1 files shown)


bettycheng14 (approved):
Good work! The fix is correct and ready to be merged.

PR #2082, opened 2026-08-16
DS: Usecase Update UC00053

jleckie89-dotcom:
This pull request addresses the review checklist for the UC00053 Air Quality Prediction use case.

Changes include:
- Fixed Folium map coordinates and popup output.
- Fixed undefined prediction/residual variables.
- Corrected MSE and R² narrative values.
- Removed unused libraries and cleaned notebook outputs.
- Applied Australian English where required.
- Applied the MOP use case template, including CSS styling, linked contents and numbered sections.
- Re-organised notebook headings and documentation.
- Added numeric validation to the interactive prediction inputs.


molliefernandez-mentor (changes requested):
Hi @jleckie89-dotcom, Some feedback for the use case!

- Delete any old files in the folder for this use case, the old .ipynb, and HTML and the JSON
- move the new file into the FINALISED folder
- update the use case index file - add to the table at the top and delete where your use case is in the other tables.
- The usecase-header-bar needs to eb applied to the use case title.
- combine the content in the scenario and user story sections and delete the user story heading
- in the 'What this use case will teach you' section delete what is there (it doesn't really make sense) and add some bullet points about what the analysis teaches you about the data, also think about what you learnt yourself when working through this use case, what did the notebook teach you?
- The background section should not be in first person, and it should be more about the whole scenario rather than what the analysis will teach you, which is what it is talking about
- delete the whole references section
- remove the bullet points that are next to the 2.1 and 2.2 sections in the contents.
- delete the sentence "Note the "Discussion" sections are displayed under relevant code outputs."
- there is no 'numbered-heading' div tag, the sub headings should be under the usecase-sub-heading div tag.
- All of your smaller sub headings that comes after the contents list should be numbered and included in the contents list. There should not be any 'usecase-unnumbered-heading' div tags after the contents list. they should all be numbered sub-headings that are listed in the contents page. 
- I also got a few errors when running the code:
Image

- You also don't need to !pip install every time, might need to find a better solution to this as it is time consuming and redundant after one go.
Image

Great start! A few things to still work on but on the right track!

jleckie89-dotcom:
This PR has been superseded by #2114 due to the updated repository structure. The reviewer feedback has been addressed and the finalised UC00053 notebook and updated Use Case Index are now included in #2114.

PR #2083, opened 2026-08-16
WD: Fix/auth getauthuser objectid

lakindubd:
Fixes two bugs that were completely blocking admin actions across the whole app, logged-in admins were getting `401 User not authenticated` on every admin-protected route (use cases, contributors, categories). Both trace back to the recent auth migration leaving part of the old Postgres-era contract in place.

Bug 1 shared auth helper rejects all admins (the big one)

`getAuthUser()` in `src/app/api/library/auth.ts` was running `Number()` on the `x-user-id` header to check authentication. Since the auth migration (PR that moved login/signup/session to MongoDB), that header carries a **Mongo ObjectId string** , not a numeric Postgres id. `Number` is `NaN`, so `isAuthenticated` was always `false`, so every request through this shared helper was rejected as unauthenticated, before the admin check even ran.

Because this one helper is shared, it broke admin auth on use cases, contributors, and categories at the same time.

**Fix:** treat `x-user-id` as the string it now is (kept as-is for downstream `findById`, which wants the string anyway). The `roleId`/`isAdmin` logic is unchanged that side was never broken.

Bug 2 use-case create rejected `created_by`

With auth fixed, creating a use case then failed with `400 created_by is required`. The create route was expecting `created_by` in the request body, but the form no longer sends a valid one.

**Fix:** the route now stamps `created_by` from the authenticated admin's own id (server-side), instead of trusting the client to supply it. This is both correct and more secure, a client can no longer attribute a use case to another user.

Files changed

- `src/app/api/library/auth.ts`  treat `x-user-id` as an ObjectId string, not a number.
- `src/app/api/usecases/route.js`  set `created_by` from the authenticated session, not the request body.

### DIFF
--- a/next_webapp/src/app/api/library/auth.ts
+++ b/next_webapp/src/app/api/library/auth.ts
@@ -5,7 +5,7 @@ export function getAuthUser(request: NextRequest) {
   const roleIdRaw = request.headers.get("x-user-role-id");
   const roleName = request.headers.get("x-user-role");
 
-  const userId = userIdRaw ? Number(userIdRaw) : null;
+  const userId = userIdRaw && userIdRaw.trim() !== "" ? userIdRaw : null;
   const roleId = roleIdRaw ? Number(roleIdRaw) : null;
 
   return {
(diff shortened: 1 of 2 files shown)


suba-t24 (approved):
Since you've already verified the fix end-to-end, the logic is sound and addresses both blockers gracefully without introducing regressions. The PR is good to go!

PR #2086, opened 2026-08-17
WD: Fix Profile dark/light mode synchronization

j0j0-ops:
## Summary

Fixes the Profile page theme synchronization issue where the page could become inconsistent with the site's selected dark/light mode.

## Changes

- Removed the Profile page's independent `darkMode` state.
- Removed the local effect that directly added/removed the `dark` class.
- Profile now follows the application's existing global theme state.

## Testing

- Tested the Profile page in dark mode.
- Navigated away from Profile and returned; dark mode remained consistent.
- Tested the same behaviour in light mode.
- Confirmed the Profile page stays synchronized with the selected site theme.

## Result

The Profile page now consistently follows the site's dark/light mode when navigating between pages.

### DIFF
--- a/next_webapp/src/app/[locale]/profile/page.jsx
+++ b/next_webapp/src/app/[locale]/profile/page.jsx
@@ -10,7 +10,6 @@ const Profile = () => {
   const router = useRouter();
   const params = useParams();
   const locale = params?.locale || "en";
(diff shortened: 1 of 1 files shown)


suba-t24 (approved):
The PR is clean, safe, and fixes the issue as described

PR #2087, opened 2026-08-17
WD; Add RTL support for Arabic locale in layout, header and footer

Ramin619:
## What this PR does
Adds right-to-left (RTL) language support for the Arabic locale.

## Changes
- **layout.tsx**: the `<html>` tag now sets `dir="rtl"` when the active locale
  is Arabic (`ar`), and `dir="ltr"` for all other languages.
- **Header.tsx**: replaced physical CSS classes (`ml-*`, `left-0`, `text-left`)
  with logical ones (`ms-*`, `start-0`, `text-start`) so the header mirrors
  automatically under RTL.
- **Footer.tsx**: same logical CSS refactor for the text alignment, and
  converted the inline-style borders/padding and the raw `<style>` block to
  logical CSS properties (`border-inline-*`, `padding-inline-*`,
  `inset-inline-start`). Also added RTL-specific rules so the quick-link hover
  nudge and arrow point the correct way in Arabic.

## Testing
- Visited `/ar` and confirmed `dir="rtl"` is applied and the header and footer
  mirror correctly (align right).
- Confirmed `/en` and other languages still display left-to-right and are
  unaffected.
- Checked dark mode and all 8 languages still work.

### DIFF
--- a/next_webapp/src/app/[locale]/layout.tsx
+++ b/next_webapp/src/app/[locale]/layout.tsx
@@ -14,7 +14,7 @@ export default async function LocaleLayout({
   const messages = await getMessages({ locale });
 
   return (
-    <html lang={locale} suppressHydrationWarning>
+    <html lang={locale} dir={locale === "ar" ? "rtl" : "ltr"} suppressHydrationWarning>
       <head>
         <script
           dangerouslySetInnerHTML={{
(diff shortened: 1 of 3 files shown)


suba-t24 (approved):
The PR correctly implements foundational RTL support using modern logical CSS properties. The acceptance criteria have been entirely met.

PR #2088, opened 2026-08-17
WD:Fix language translations across pages

j0j0-ops:
## Summary
Fixed and expanded multilingual support across the MOP website.

## Changes
- Added translations for the Use Cases page
- Translated search, reset, pagination and use case UI text
- Added translations for the About page
- Added translations for footer content
- Updated translation files for all supported languages
- Added CSS TypeScript declaration support
- Maintained dark/light theme compatibility

## Testing
- Tested language switching locally
- Verified Use Cases page translations
- Verified About page translations
- Verified pagination and footer translations
- Tested in both light and dark mode


champa1n73 (changes requested):
### 1. Content search is currently broken.
searchbar.tsx exposes "content" as an option, but /api/usecases only accepts "title", "description", and "all". Selecting Search by content sends search_by=content and receives a 400 response. Please remove this option or update the API contract.

### 2. Category data is lost in the use-case cards.
The API now returns the embedded category object, but the client still reads u.category_id, which is no longer present in the DTO. Please update the mapper to use u.category or remove the unused field.

### 3. The initial search mode is not validated.
searchParams.get("search_by") is cast directly to LocalSearchMode. Invalid URLs can set an unsupported mode and cause the API request to fail. Please validate the value against the supported modes and default to "title" when invalid.

### 4. Two translations are incorrect.
In both Chinese and Spanish, "Search by content" is translated as "Search by tag". This gives users the wrong search-mode label.

### 5. The API response ID and TypeScript type disagree.
toUseCaseDTO() returns id as a string, while CaseStudy.id is declared as a number. Since the response is mapped through any, TypeScript does not catch this mismatch. Please either update the interface to string or normalize the ID consistently.

### 6. The About stylesheet is global despite being co-located.
about.css defines body { font-family: "Montserrat", sans-serif; }, so importing it from the About page changes the font for the entire application whenever that stylesheet is loaded. Please scope the styles to the About page or move shared font styling to the global stylesheet.

### 7. The footer still contains untranslated user-facing text.
"Melbourne Open Data" and the external-link indicator are hard-coded. If the goal is complete multilingual support, the link label should use a translation key as well.

### 8. Footer sections lack semantic navigation structure.
The Quick Links and Connect labels use <p> elements, and the related links are not grouped in <nav> elements. Using headings with aria-labelledby would improve screen-reader navigation.

### 9. The footer is not fully RTL-friendly.
It uses physical properties such as left, border-left, padding-left, and text-left. Arabic locales may render these in the wrong direction. Prefer logical properties such as inset-inline-start, border-inline-start, padding-inline-start, and text-start.

champa1n73:
The positive parts of this PR are:

- The separation between getTranslations() for the server-rendered About page and useTranslations() for client components is appropriate.
- Moving the About styles next to the page improves discoverability and ownership.
- The translation keys cover the major user-facing sections, including empty states, pagination, newsletter validation, and search controls.
- The footer newsletter validation and toast messages now respect the active locale.
- The use-case preview has improved structure, clearer empty states, and better visual hierarchy.
- The pagination component retains a shared implementation across public and admin views while translating its labels.
- The existing RTL handling in parts of the footer, such as [dir="rtl"] link movement, is a useful foundation, but it should be applied consistently.

champa1n73 (approved):
All problems are resolved.

PR #2089, opened 2026-08-17
WD: test: rewrite Jest and Cypress tests for MongoDB

Kangbunbun:
## Summary

- Rewrote Jest coverage for the migrated MongoDB login, profile, and use-case routes.
- Restored the POST use-case suite as a discoverable TypeScript test.
- Updated the Footer test to match the current component.
- Replaced stale Cypress specs with live Atlas API smoke tests.
- Separated Jest and Cypress TypeScript scopes.
- Removed obsolete Cypress tests that referenced old routes or were not discovered.

## Validation

- Jest: 13 suites passed, 143 tests passed
- Cypress: 1 spec passed, 5 tests passed against Atlas
- Cypress TypeScript check passed
- Production build completed successfully (34/34 static pages)
- Credential and whitespace checks passed

## Additional findings

- The legacy navigation tests exposed a React hydration mismatch on `/en`. This was not suppressed and should be investigated separately by the frontend team.
- Root TypeScript checking now runs, but still reports existing errors in unchanged files.
- The existing ESLint compatibility warning remains during build, but does not stop the build.


suba-t24 (approved):
The PR removes all legacy Supabase traces from the modified test files and aligns the test suite with the MongoDB reality. Great work!

PR #2090, opened 2026-08-17
Git: Cleaned up the old .ARCHIVE folder

maanak-16:
Went through `.ARCHIVE/` as requested.

* **Moved:** `usecase-pedestriantrafficanalysis.ipynb` into the current work and used it to populate the previously empty **UC00019** notebook (Update Needed).
* **Deleted:** All of `.ARCHIVE/MOP` and the legacy code, dashboards, and websites in `.ARCHIVE/prior_work` that are already covered by newer use cases.
* **Kept for confirmation:** Historical datasets in `prior_work` containing 2020–2021 sensor and weather data that cannot be reproduced through live APIs. Please confirm whether these should also be removed.
* **Not touched:** `.ARCHIVE/ETL`, as it contains AWS Lambda deployment files. I wanted to confirm whether it is still active before removing it.


bettycheng14:
Thanks @maanak-16 for the clean up! 
Confirmed that `.ARCHIVE/ETL` is not referenced anywhere in `next_webapp`. There are no dependency or references to `ETL`, `parking_availability`, `PlaygroundETL`, or `AWS` services in the `next_webapp` directory.

Also, the Web Dev team is using Next.js + MongoDB Atlas + GCP and no longer uses `AWS Lambda` or `SAM`. Therefore, this folder has no dependency on our current stack, ready to clean from our side.

molliefernandez-mentor (approved):
Looks good to me, thanks @maanak-16, wait for @bettycheng14 to approve and confirm about the Lambda files you mentioned and then happy for you to merge!

bettycheng14 (approved):

maanak-16:
Removed the ETL folder and the three raw dataset folders as well, now that both were confirmed clear. .ARCHIVE is fully cleaned up on the DS side.

PR #2096, opened 2026-08-19
WD: feat: added notebook download button on use case page

Muhib1-lewis:


suba-t24 (changes requested):
Overall, the implementation looks good and correctly utilizes the existing API endpoint to trigger a download!

However, there's one edge case we should handle: Some of our legacy use cases are stored as HTML rather than Jupyter Notebooks (useCase.content_type === "html"). The current implementation hardcodes the .ipynb extension in the download attribute, which means if a user downloads an HTML use case, it will be saved with a .ipynb extension and result in a corrupt/invalid file.

Could you update it to dynamically set the extension and button text based on the content_type?

suba-t24 (approved):
Since HTML legacy files are being removed, approving the PR

PR #2097, opened 2026-08-19
DS: Usecase Update UC00139

whysolazyru:


Ruki-Diaz (changes requested):
Ok there are a few changes are required in this,

* Starting off with the main task, add the CSS styling code chunk from the current use case template at the very beginning of the notebook, before the use case content.
* Please update all main section and subsection headings to follow the exact heading structure and formatting of the current use case template.
* Please ensure the Contents section and Content Divider follow the current template.
* Please add a proper Conclusion section following the current template structure, including “What We Achieved” , “What We Learned”, and "Observations for further opportunities "
* Please clean up unnecessary standalone/instructional text and ensure explanations are placed under the appropriate sections.
* Make sure all visualisations follow the colour palette and colour order specified in the current use case template. In particular, please review the final MOP accessibility map, as it currently uses default plotting colours rather than the template colours.

ashleyyy01 (changes requested):
Rukshan covered most of it. Just a few more things to add on:

- Add a "Data Sets" section near the top listing all datasets used, with the links used 
- K-Means is imported in the library list but never actually used anywhere in the notebook so  either implement it to match "What You Will Learn," or update that section to reflect the rule-based scoring method actually used
- Same for Folium - it's imported but no interactive map is ever created
- There are also a few typos and some US spellings left in the notebook, so please go through them once as well.

One small thing theres "link_microclimate = '' defined but empty, and doesn't seem to be used anywhere else. Not sure what it's for... leftover from earlier? could you confirm if it's needed, or should it just be removed?

Thank you! Awaiting your changes.

molliefernandez-mentor:
@whysolazyru looks like there is a lot of work still to be done on this use case, I would recommend looking at other completed use cases for reference and using the template. 
Once you have applied the changes suggested by the other reviewers I will do a proper review. 
Please apply the changes suggested.

molliefernandez-mentor (changes requested):
Thanks @whysolazyru 

- Please put the use case in the FINALISED folder. You should not be using a Playground folder with a different students name anyway.
- Please update the use case Index file
- Please delete any other files for this use case in the READY TO PUBLISH folder
- change the date created to the trimester and year that Ananya created the use case, not this trimester.
- remove the "UC00139:" from the title of the use case
- The headings should be in separate text boxes to the body text, give them more space between
- Please update the scenario so that it reflects who the target audience would be for this usecase not what your role is in analysing this data. below is an example:
Image

- delete the second paragraphh in the scenario, move the first paragraph to the background section and then come up with a scenario for this use case. 
- move the sentence for datasets to above the list of the datasets. 
- Make sure all of your datasets listed have their full names (currently they are not the full name of the dataset. 
- the contents sub section should be 2.1, 2.2 etc. not A,B,C,D
- remove "step" from the start of each of the numbered headings, just keep the numbers
- In the code commenting there are 'steps' that don't match the headings you have included. Can you please remove all of these steps in the code commenting throughout the use case. (makes it look like the whole thing was written by AI)
- after every numbered heading you have a paragraph that states what this section summarises or interprets and this is not necessary after every single section, especially when the second paragraph tells you what you want to know. I would merge the 2 paragraphs under each heading. and in some cases just delete the "this section summarises..." paragraph all together 
- 5.5 and 5.6 should be moved under "conclusion" the key findings can be merged with the "what we learnt" and limitations can be added. all the headings under the conclusion should not be heading div tags, they should be bolded body text. 

Still a lot of work to be done :)

Ruki-Diaz (commented):
Once the above changes are addressed, I’ll be happy to review the use case again.
If anything in the feedback is unclear or you need any help making the changes, please feel free to reach out. I’m happy to help!

whysolazyru:
Hi Mollie, I’ve addressed the requested changes and uploaded them to this existing PR. The use case has been moved to the FINALISED folder, the Use Case Index has been updated, old copies have been removed, and the requested notebook changes have been applied. Thank you!

molliefernandez-mentor (changes requested):
Hi @whysolazyru 

- please update the use case index file
**Author**: {Author/s}
**Level**: {Intermediate}
- The "1.import required libraries" heading in the contents page does not exist 
- the 5.4 Section has some sort of formatting issue
Image

- in the conclusion section above the sub headings can you please write a short paragraph which outlines how the problem statement talked about in the scenario has been achieved.
- Most of the use case is spent cleaning the many datasets included in this use case, I wonder whether there could be additional visualisations added?

molliefernandez-mentor (changes requested):
Thanks @whysolazyru 

- in the use case index can you please put your use case in order of the UC code. 
- Can you please remove all the breaks in the conclusion section, 1 break between sections is enough.

- I am getting this error when running the code. 

Image

molliefernandez-mentor (approved):
Thanks @whysolazyru, just a small thing, please change T1 and T2 to Trimester 1 and Trimester 2. 

Other than that it looks good, please wait for @ashleyyy01 and @Ruki-Diaz 's review then you can merge.

PR #2101, opened 2026-08-21
WD: fix: improve footer navigation landmark labels

himeshsam93:
Updated the Footer navigation landmarks based on accessibility review feedback.
Linked the Quick Links navigation to footer-quick-links-heading using aria-labelledby.
Linked the Connect navigation to footer-connect-heading using aria-labelledby.
Added semantic heading IDs to ensure each navigation landmark has a distinct accessible name.
Kept the changes limited to Footer.tsx to align with the current master branch.
This prevents screen readers from announcing both Footer navigation landmarks with the same name.


suba-t24 (changes requested):
Great initiative adding semantic nav tags and aria-labelledby for accessibility! Before merging, please address these three quick issues:
- Heading Hierarchy: Change the h3 for "Connect" to an h2. Since "Connect" and "Quick Links" are sibling columns in the footer grid, they should be at the same heading level.
- Missing Column: The 4th column (Newsletter) was skipped. Please apply the same h2 treatment there so the footer has a consistent document outline.
- i18n Compatibility (text-start vs text-left): Please revert the text-left classes back to text-start. This app uses next-intl, and logical properties like text-start are required to ensure the footer aligns correctly if viewed in a Right-To-Left (RTL) language.

suba-t24 (approved):
Everything looks excellent. Great work on improving the accessibility and structure of the footer!

PR #2104, opened 2026-08-21
DS: Usecase Update UC00111

AtishayDeakin:
Renamed `Real-time Parking Hotspots` → `Melbourne Parking Hotspot Analysis`.

Bug fixes
- Vectorised the `for` loop + `pd.concat` anti-pattern in section 4.5 — verified identical output to the original loop
- Fixed `SettingWithCopyWarning` on `todays_data` (now created with `.copy()`)
- Fixed the "Today's Date" filter — previously left every downstream cell empty if the feed had no records for the current date; now falls back to the most recent available date
- Added `na=False` to `str.contains('DIS')`

Template alignment
- Restructured to the current template: header bar, metadata, unnumbered front matter, 8 numbered sections, clickable Contents page
- Removed deprecated classes; added missing Background, Data Sets and References sections

Content and cleanup
- Added narrative context between code blocks
- Removed commented-out Colab/API key code, unused imports and `{}` placeholders; fixed Scenario grammar
- Applied the colour palette across charts and added map legends

Testing: Full Restart & Run All — 49 cells, sequential, no errors or warnings.


molliefernandez-mentor (changes requested):
Hi @AtishayDeakin, thanks for submitting this, mostly just small changes :)

- You are not using the most current version of CSS, this version has sections that are no longer in use. Please copy the latest version from the template.
- Maybe add a sentence or something leading into the 2 listed questions in the scenario they are kind of just floating there. something like "This use case will provide live answers to the following questions" or something like that.
- Remove the references section.
- in the Scenario you talk about "Parking Peaks" but the name of the use case and throughout the use case you use "hotspots" can you please pick one word to use throughout and keep it consistent. And whichever word you use please define what it means in the scenario. 
- Please update the name of the use case to be: "Melbourne Live Parking Availability"
- Please delete the old entry of US00111 in the use case index under the T2 2024 table.

AtishayDeakin (commented):
Hey @molliefernandez-mentor, I have updated the PR with the changes you mentioned, please have a look at it. Thanks.

molliefernandez-mentor (approved):
Thanks @AtishayDeakin looks good. 
Please wait for @VedS8's review and then you can merge

VedS8 (approved):
Great work Atishay, looks good to me!

PR #2105, opened 2026-08-23
WD: fix: migrate password reset flow to MongoDB

CHs-project-max:
migrate the password resetign rountes to use mongoDB instead of old supabase


HarshanaDeakinUni (approved):
LGTM, original implementations is ok. additinally, there was a small bug in the original code that was related to SMTP implementation, which was mitigated with this implementation. Good work!!!

lakindubd (approved):
bug has been addressed! lgtm!

PR #2109, opened 2026-08-26
DS: Usecase Update UC00081

Hariharan1202:
- Renamed and aligned the use case name across the notebook title and metadata
- Reformatted the notebook onto the current template (CSS injection cell, current section classes, numbered Contents with working anchors, use case datasets box with dataset citations)
- Fixed the typos and trimmed the References section to genuinely authoritative sources
- Scoped the warnings suppression, summarized the ARIMA grid-search output instead of printing per failure, and removed redundant print loops that duplicated chart data
- Added a plain-language explanation of ARIMA before the modelling cell, plus explicit notes on the small sample size
- Fixed a pandas compatibility bug (freq='A', removed in pandas 3.0) and added static fallbacks for the two interactive widget charts so they render outside a live kernel


molliefernandez-mentor (changes requested):
Hi @Hariharan1202,

- Please move the file into the FINALISED folder.
- Please update the use case index file.
- Update your CSS, it is an old version
**Author**: {Author/s}
**Level**: {Intermediate}
- Add the Trimester that the use case was created
- delete the line about combining datasets under the datasets section
- The second dataset has not got its full name. 
- the limitations should eb included in the conclusion and should not be in the contents
- the conclusion should also not be in the contents (it is an un numbered section.
- Delete the references section
- In the whole use case there needs to be more explanation of what is happening and why you are doing things, each section should at a minimum have a paragraph explaining what is going on and why each step is taking place, this is a tutorial. 
- The conclusion needs to have the sub sections that are listed in the use case template - the limitations can be one of these sub headings under the conclusion. 
- If there are any references they should form part of the explanation of that specific section and where additional information can be found.

liajohn30 (commented):
Hi Hariharan, I’ve reviewed the use case as well. The additional explanations have made the overall flow much clearer. I agree with Mollie’s feedback, and there are still a few of those points to address. Just from my side, I’d also suggest making sure the conclusion briefly brings together the key findings from the analysis before the final recommendation. Once these are addressed, it should be good to go. Thanks!

Hariharan1202:
Hi @molliefernandez-mentor and @liajohn30 , I have made the requested changes and rebased the branch onto the latest master.

- Moved the use case into the FINALISED folder and updated the use case index
- Updated the CSS to the current template and the use case details to follow the template
- Added the trimester the use case was created
- Deleted the line about combining datasets and added the full name of the second dataset
- Removed the References section and included the links within the sections they support
- Added an explanation to every section of what is happening and why
- Made the Conclusion an unnumbered section that is not in the contents, added the template sub sections with Limitations as one of them, and added a summary of the key findings before the recommendation

Please let me know if anything else is needed.

molliefernandez-mentor (changes requested):
Hi @Hariharan1202 everything looks great but the folium maps have broken, this is a common issue we have seen across all use cases, message the DS chat and see how other students have solved the issue:

Image

Please make sure that all of your visualisations are using the colours from the template.

PR #2110, opened 2026-08-26
DS: UseCase Update UC00213

Hariharan1202:
- Updated the report to match the new template, including the HTML block, Data Sets section, conclusion, and section numbering.
- Added the Date Created field and a hyperlinked table of contents.
- Updated plot colours to follow the required colour guidelines.
- Removed references, citations, and unnecessary JSON/HTML files.
- Updated the user story to the required format.


NguyenMav (dismissed):
Hey @Hariharan1202, I noticed that you completed all the tasks in the planner card. Awesome work! There's still a lot of work left to be done since the time I audited that use case, as there have been a lot of changes and the FINALISED use cases are very specific about what they want.

---

Here is the list of things that need to be updated in the notebook from what I can see:
- Update the CSS at the top of the notebook to match the one in the use case template that has been recently updated.
- Change "Prerequisite Skills:" to "Pre-requisite Skills:" just to match the template.
- Change the "Introduction" header to "Background" to match the template exactly.
- It seems there's been a change in the results/numbers, so try to make sure the numbers in the description match the outputs. Though I believe the use case is using data that is being updated in real-time, so another option is to avoid using numbers that are changing in the description and use comparative language. Up to you. But here's an example from the numbered section "3.8. Checking The ID Columns", where the description states 137 unique locations, but the output is 134.
- Bullet lists in the description need to be wrapped in the "usecase-list" for 2.1, 2.2, 2.3, 6.3, 6.4, 10.3, 10.4 and 10.7. Similar to what you've already done with "What this use case will teach you".
- Make sure the plots are following the colour palettes specified in the template. Like "10.9. Actual vs Predicted Visualisation", use "color="black")" in the code block, which is not part of the colour palette specified in the template.
- In the description of "1. Importing The Libraries", it said "A random seed was set with a student ID to allow reproducibility of the outputs in this notebook", but there was no seed in the code block below it, until "10.1. Setting Random Seeds for Reproducibility". So don't forget to check over the descriptions too. Also, please swap it with a random number to avoid showing a student ID, as this will be published publicly.
- Add the id attributes to the unnumbered headings, like id="scenario", id="background", id="conclusion", etc.
- Let's have the "Conclusion" as a separate section, since the template requires "What we achieved in this analysis", "What we learned from this analysis", and "Observations for further opportunities" parts. So don't forget to add that in!

---

Here are the things that are more admin-related:
- So move the use case notebook file to the FINALISED folder (so it should be removed from the READY TO PUBLISH folder).
- In the Use_Case_Index.md, I want you to update the "Finalised" table at the very top with this use case. (I couldn't find the use case in the "Ready To Publish" tables, specifically, "T1 2026", so you can skip the part of having to remove an entry.)

---

It's easiest to have a reference, so check out the completed use cases in the FINALISED folder that I based my review on (I highly recommend going over one of the use cases and checking each individual cell to see how this current use case can be further improved): https://github.com/Chameleon-company/MOP-Code/tree/master/usecases/FINALISED

Also, refer back to the template: https://github.com/Chameleon-company/MOP-Code/blob/master/usecases/usecase_TEMPLATE.ipynb

molliefernandez-mentor:
@Hariharan1202 please make the changes to the PR ASAP

molliefernandez-mentor:
@NguyenMav you cannot be quality assuring a use case you authored, it is a conflict of interest. I told you at the start of the trimester Quality managers should start from the top of the list (lowest UC numbers first) so that no use case was worked on that had been created by a current student. I am removing you from this and adding a different quality manager. 

@Hariharan1202 please apply the changes @NguyenMav has suggested but I will assign a new reviewer for the final approvals.

molliefernandez-mentor (changes requested):
Hi @Hariharan1202 

- Update the use case index
**Author**: {Author/s}
**Level**: {Intermediate}
- Change "Introduction" to "Background"
- remove the sentence under the dataset heading, it is unnecessary
- the conclusion should be an unnumbered heading and does not need to be in the contents list. 
- the conclusion should have the sections listed in the template. 
- Please make sure that the conclusion addresses the problem defined int he scenario and talks about how the use case has addressed it.

Hariharan1202:
Hi @molliefernandez-mentor,  I have applied the changes suggested by NguyenMav along with your feedback, and rebased the branch onto the latest master.

- Moved the use case into the FINALISED folder and added it to the Finalised table in the use case index
- Updated the CSS to the current template, updated the use case details to follow the template, and renamed "Introduction" to "Background"
- Removed the sentence under the Data Sets heading
- Re-ran the notebook so the outputs are current, and updated the descriptions to match them. Where numbers change with the live data, I used comparative wording instead
- Wrapped the bullet lists in the usecase-list class, and added id attributes to the unnumbered headings
- Updated the plots to use the template colour palette
- Corrected the random seed description and replaced the student ID with a random number
- Made the Conclusion an unnumbered section that is not in the contents, with the template sub sections, and made it address the scenario

Please let me know if anything else is needed.

molliefernandez-mentor (approved):
Thanks @Hariharan1202 

Looks great, all good from me. Please wait for @snimle123's review then you can merge.

snimle123 (approved):
Hi @Hariharan1202, I’ve completed my review and everything looks good. Approved.

PR #2115, opened 2026-08-31
WD: global API error handling with apiFetch, toast notifications and ErrorBoundary

Muhib1-lewis:
Implemented global API error handling for Sprint 3.

- Added apiFetch() a fetch() wrapper that catches failures and 
  shows a toast notification (sonner). Used instead of Axios 
  interceptors since the app barely uses Axios.
- Added ErrorBoundary wired around NotebookRenderer to catch 
  render-time crashes.
- Migrated 39 files (read-only pages, admin CRUD, auth flows) to use 
  apiFetch instead of raw fetch() so failures no longer fail silently.
- Auth and contact forms use silent mode since they already show 
  their own error messages so  those will stay unchanged.
- Fixed one pre-existing bug along the way: profile/page.jsx only 
  showed errors for "Unauthorised" failures before now shows errors 
  for any failure.

### DIFF
--- a/next_webapp/package-lock.json
+++ b/next_webapp/package-lock.json
@@ -37,6 +37,7 @@
         "react-tag-input-component": "^2.0.2",
         "rehype-raw": "^7.0.0",
         "remark-gfm": "^4.0.1",
+        "sonner": "^2.0.8",
         "winston": "^3.19.0"
       },
       "devDependencies": {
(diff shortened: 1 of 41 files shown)


suba-t24 (approved):
Great work on this PR! The implementation perfectly addresses the requirements for handling frontend API errors and protecting the application state.

suba-t24:
@Muhib1-lewis, I’ve approved the PR, but I noticed a couple of minor things that we can clean up (either in this PR before merging or in a quick follow-up):
Minor Frontend Suggestions:
Extend apiFetch for Non-JSON responses: In src/app/[locale]/usecases/[id]/page.tsx, there's a bare fetch left because it requests /content (which returns HTML). You could extend apiFetch to accept a responseType: 'json' | 'text' option so we can utilize it there as well.
Dashboard Live Search: In src/components/Dashboard.tsx, the debounced live search still uses a bare fetch wrapped in a try/catch to suppress errors. You can migrate this to apiFetch(url, { silent: true }). This will maintain consistent error formatting while preventing toasts from popping up every time a user types a typo.
Uploads (Just an observation): I noticed src/app/[locale]/upload/page.tsx still uses axios. This is absolutely the right approach since fetch doesn't natively support tracking upload progress without ReadableStream boilerplate. The local error handling there looks good.

PR #2118, opened 2026-09-01
WD:feat(admin): make admin panel responsive with mobile card tables and …

abdullahsher445:
## Summary
This PR addresses **QM Feedback (Responsive Admin Panel)** by eliminating horizontal scrolling across mobile viewports (<768px) and refactoring the administrative layout for smooth mobile usability.

---

### 1. Responsive Card Table View (<768px)
Refactored administrative data tables so each record transforms into a vertically stacked card on mobile screens, displaying full record metadata with immediately accessible **Edit** and **Delete** action buttons (eliminating horizontal swiping):
- **Use Cases**: `UseCaseTable.tsx`
- **Blogs**: `BlogsTable.tsx`
- **Categories**: `categories/page.tsx`
- **Contributors**: `contributors/page.tsx` (eliminates the desktop 1000px table overflow)
- **Activity History & Recent Activity**: `activity-history/page.tsx` & `AdminRecentActivity.tsx`

*Note: Full tabular layouts and column structures remain intact for desktop viewports (`>= 768px`).*

### 2. Mobile Navigation Drawer & Header
- **Off-Canvas Drawer**: Configured `AdminSidebar.tsx` to remain hidden off-canvas on mobile (`< 1024px`), eliminating the fixed green block/floating icon overlays.
- **Drawer Controls**: Tapping the header hamburger icon (`☰`) smoothly slides out the navigation drawer over a darkened backdrop. Includes an `✕` close button and auto-closes on route changes.
- **Collapsible Desktop Sidebar**: Preserved collapsible desktop view (`w-[190px]` expanded / `w-[70px]` collapsed).

### 3. Layout & Viewport Overflow Fixes
- **Responsive Page Headers**: Updated headers across all admin pages (`Use Cases`, `Categories`, `Blogs`, `Contributors`, `Gallery`, `Dashboard`) to wrap cleanly (`flex-col sm:flex-row`) with fluid typography.
- **Removed Conflicting Margins**: Cleared hardcoded left margin `ml-[88px]` in Activity History and negative margin `-m-8` in Gallery that previously caused viewport shift.
- **Bounded Layout**: Added `overflow-x-hidden` and `w-full` constraints to `AdminLayout.tsx` to prevent horizontal screen shift and exposure of background gutters.

---

## Before
01_before_categories_landscape
02_before_use_cases_landscape
03_before_dashboard_sidebar_landscape
04_before_dashboard_landscape

## After
06_after_sidebar_landscape
07_after_use_cases_landscape
08_after_categories_landscape
05_after_dashboard_landscape


hamizhassan16 (changes requested):
Thanks for this, the responsive pattern is applied consistently across all the admin pages and the breakpoints line up with what's described. A couple of things I'd like addressed before merging:

1. Escape key doesn't actually close the mobile drawer
AdminSidebar.tsx — the backdrop's onKeyDown only fires when the backdrop div itself has focus:

<div role="button" tabIndex={0} onClick={closeMobileMenu} onKeyDown={(e) => e.key === "Escape" && closeMobileMenu()} ... />

Since nothing calls .focus() on it when the drawer opens, pressing Escape while focus is on a nav link (or anywhere else on the page) won't close the menu. Can you move this to a document-level keydown listener in a useEffect (active only while mobileMenuOpen is true) so Escape works regardless of what's focused?

2. any typing introduced in BlogsTable
BlogsTable.tsx — the new mobile card block uses data.map((item: any) => (, while the existing desktop table right below it relies on the inferred/typed prop with no any. Can we type this consistently instead of widening to any?

3. Unrelated change in profile/page.jsx
This file's diff is a pure template-literal reformat (moving ?/: onto different lines) with no functional difference, and it's unrelated to the admin panel work this PR is about. Can you either drop this from the PR or explain why it's needed here? Prefer keeping the diff scoped to the stated change.

Everything else looks good — happy to approve once these are addressed.

hamizhassan16:
Also check if PR#2112 is merged before you merge this as there might be some conflicts

hamizhassan16 (changes requested):
Confirmed the escape-key handler and BlogsTable typing are both fixed — the useEffect/document-listener approach and the new BlogItem interface both look correct.

Still outstanding: profile/page.jsx still has the unrelated template-literal reformat, unchanged from before. Can you drop that from the PR, or let us know why it's needed here? Once that's resolved I'm happy to approve.

hamizhassan16 (approved):
Confirmed the last item is resolved — the profile/page.jsx diff is fully reverted, diffed it against master myself and it's a clean zero. All three things I flagged (Escape key, BlogsTable typing, unrelated file) are addressed. Approving.

PR #2121, opened 2026-09-02
WD: Clean up Playground root README

hamizhassan16:
Playground had two READMEs sitting at its root, both of which were actually leftover write-ups for unrelated individual projects, not documentation for the folder itself.

Replaced both with a single README explaining what Playground is actually for: a working space for in-progress use cases and rough/draft datasets. Also notes that once a use case gets published to usecases/READY TO PUBLISH, the student should delete their folder from Playground to keep it clean.


molliefernandez-mentor (changes requested):
Under the 'what goes here' section can you add a part about each student creating their own folder that they can work in. 

Can you add a bit more about how students should delete their playground folder at the end of every trimester. And that junior students can keep their folder between trimesters but seniors must delete their playground folder. 

Also Add the chameleon logo like you have in the main readme.

hamizhassan16:
ok I will do that

hamizhassan16:
@molliefernandez-mentor Kindly check now

molliefernandez-mentor (approved):
Thanks, looks great!

PR #2124, opened 2026-09-03
DS: Usecase Update UC00070

sooryaravi24:
Random Forest models now use meaningful predictors (CLUE location area, industry type, seating type, and each property's historical capacity) instead of the arbitrary property_id and report R² alongside MSE.

Replaced the input() call in the prediction step with an ipywidgets dropdown, so the notebook no longer halts execution and renders correctly as static HTML.

ARIMA forecasting now flags properties with too little history to produce a meaningful trend and reports a one-step holdout accuracy check.

Replaced the hardcoded 13-address to coordinate dictionary with a CSV lookup file (missing_address_coordinates.csv) kept alongside the notebook, and fixed the deprecated fillna(method="ffill") syntax.

Reduced redundant head()/info() dumps and consolidated repeated warnings.filterwarnings() calls.

Rebuilt to the current MOP template: CSS styling, numbered headings with descriptions, a Data Sets section with links, a Contents page, and Australian English spelling throughout. Removed the References section.


molliefernandez-mentor (changes requested):
thanks @sooryaravi24,

- please move the use case into the FINLIASED folder
- please update the use case index file
- Any dependencies need to be in a folder in the DEPENDENCY folder
- Please change the name of the use case to "Hospitality Venue Capacity Forecasting" - so update the file name and the title within the use case
- your CSS is not the current version, please update it.
**Author**: {Author/s}
**Level**: {Intermediate}
- change the heading to "Scenario"
- delete the first sentence under scenario
- project goal is not a heading included in the template, please adjust it so that the content fits under one of the heading listed in the template.
- under datasets please link the name of the dataset. have a look at the previously completed use cases to see the format. 
- there is still a lot of US English please update all of the use case to have Australian English (Initialization, in the contents should be initialisation)
- delete the "The code and analysis for each section listed above begins below."
- There is not enough explanation through out the use case, this is mean to be a tutorial and so there needs to eb more explanation as to what is happening at each stage. 
- please follow the conclusion layout in the template. - the conclusion should be the last thing at the end of the use case. 
- you should not be making recommendations, this is a tutorial on how to use the data, there should not be recommendations. 
- the final impact section needs to be subsumed into the conclusion. 


still lots to do on this use case

s224251967-hub (changes requested):
Hi @sooryaravi24, I noticed a few additional points that may be worth addressing:

- Add the two mandatory template sections: What this use case will teach you and Background

- There is still one broken link in the Contents section: 1.1 Install Required Libraries (Optional) does not have a corresponding anchor.

- The exact MSE values in the Conclusion may vary a bit between every runs. As this is a tutorial, including a fixed performance values could potentially confuse readers

sooryaravi24:
Thanks for the review. I've addressed all the points kindly let me know if any further changes required

s224251967-hub (changes requested):
Hi @sooryaravi24 
Please update all visualisations to use the MOP template colour palette. For example, the coolwarm heatmap, red regression line, Set3 boxplot and default Seaborn colours do not follow the required palette.

The Conclusion link works correctly, but it is not formatted consistently with the numbered Contents list. 

A recommend: replacing freq='Y' with freq='YE' in both ARIMA cells. While 'Y' may still work in older pandas versions, it is no longer supported in current versions, whereas 'YE' is compatible going forward.

molliefernandez-mentor (changes requested):
Hi @sooryaravi24 

- Please make sure you are using only the colours listed in the template for all of your visualisations. 
- I cannot currently run the code as the dependent file is not in the location it is being called from (obviously because the PR has not been merged). 
But I don't see why you need this additional dataset of 'missing values' the values are there and I can't see any reason why they would not be pulled using the API. 
Can you please find a solution for this problem that does not involve essentially hard coding a file in the dependency folder.

PR #2126, opened 2026-09-03
WD: fix auth regression and activity logging

CHs-project-max:
Task 1 – Auth regression

Replaced local Number(x-user-id) authentication with the shared getAuthUser() helper across the affected routes.
Updated admin checks to use the shared authentication logic.
Updated profile/password to read and write passwords using the MongoDB User model instead of the old Supabase user table.
Preserved existing Supabase storage/database behaviour outside the assigned migration scope.

Task 6 – Activity history / logging

Passed the authenticated userId into relevant admin mutation error/logging paths.
Updated errorResponse() to support MongoDB user IDs.
Migrated the Activity History read route to the MongoDB Log model.
Preserved the existing Activity History filtering behaviour.
Escaped user search input before using it in MongoDB regex queries.

Testing

Project builds successfully.
Verified MongoDB error logs can contain authenticated user IDs.
Verified Activity History correctly displays valid MongoDB middleware log records.
Existing Edge middleware only persists request logs to the console; automatic persistence of successful middleware activity to MongoDB is outside the scope of this change.

### DIFF
--- a/next_webapp/src/app/api/admin/activity-history/route.ts
+++ b/next_webapp/src/app/api/admin/activity-history/route.ts
@@ -1,6 +1,12 @@
 import { NextRequest, NextResponse } from 'next/server';
-import { supabase } from '@/library/supabaseClient';
+import dbConnect from "@/lib/dbConnect";
+import Log from "@/models/mongoose/Log";
+import User from "@/models/mongoose/User";
 import { getAuthUser } from '@/app/api/library/auth';
+
+function escapeRegex(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
 
 function formatActivity(method: string | null, url: string | null): string {
   if (!method || !url) return 'Unknown action';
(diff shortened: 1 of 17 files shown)


hamizhassan16 (changes requested):
Nice consolidation overall, Two things I'd like addressed before merging:

1. There's a squished/malformed line in the activity-history route:
   `.trim();userNameMap[id] =name ||user.email || \`User #${id}\`;`
   Still valid JS, but it looks like a formatter pass got skipped. Can you run prettier/lint over the changed files before merge?

2. The old isAdmin() check accepted role name "admin" OR roleId === "1", but the new consolidated getAuthUser() only checks roleId === 1. This is probably fine since middleware sets both from the same JWT payload, but can you confirm there's no path where role name and roleId could diverge? Would rather have that confirmed explicitly than assumed.

lakindubd (changes requested):
#task1 1 - no issues all requirements met
task 6 - needs one change before merge:
the usecases routes (usecases/route.js and usecases/[id]/route.js aernt changed.  The task called "usecases, categories, contributors," and use-case create/edit/delete still call errorResponse without the userId, so those mutations won't be attributable to a user in Activity History. Can you wire the same getAuthUser().userId passthrough into the errorResponse calls on those two files?

On Hamiz's point The isAdmin narrowing, I checked this. auth.ts is untouched, so getAuthUser()'s roleId === 1 is the same admin check already used by usecases/contributors/categories and working in production since that fix merged, so this PR is bringing these 6 routes in line with the rest of the app, not inventing a new admin definition, which resolves most of the concern. To fully close Hamiz's point can you just confirm at the login route that roleId and roleName are set from the same Role lookup so roleId === 1 always means the admin role?

CHs-project-max:
roleId and roleName are both derived from the same embedded userData.role object. roleId is taken from roleData.legacy_id and converted to a number, while roleName comes from roleData.role_name, and both are written into the same JWT payload. Therefore the shared getAuthUser() check of roleId === 1 is using the same role record that supplies the "admin" role name. This PR is aligning these routes with the shared admin check already used elsewhere rather than introducing a separate admin definition.

lakindubd (approved):
all addressed LGTM! good work

hamizhassan16 (approved):
Confirmed both items are addressed ,the squished line is properly split now, and thanks for the explanation on roleId/roleName both coming from the same userData.role object, that answers my concern. Approving.

PR #2127, opened 2026-09-03
DS: Usecase Update UC00040

janzen63122-debug:
Reformat UC00040 to latest template, fix execution errors, make regression reproducible, and restructure sections.


janzen63122-debug:
Hi @molliefernandez-mentor  , I have updated another usecase UC00040.

Summary of changes:

CSS updated to the latest template version
Title now uses usecase-header-bar (title only, not the details)
Every text section is wrapped in usecase-body
Datasets listed once, each name hyperlinked with a short description
Contents displayed vertically, with subsections numbered (e.g. 4.1, 6.2)
Visualisations re-coloured to the template
Removed template prompts
I've also fixed the execution errors, made the regression reproducible, and tidied up the section structure. Only the .ipynb is in the FINALISED folder.

Ready for re-review when you get a chance, thank you.

molliefernandez-mentor (changes requested):
thanks @janzen63122-debug 

- please update the use case index file.
- some of your visualisations do not show, the ones that are "fig.show()"
- also for the folium maps where you are using the icons, please use the template colours for them.
- Some of the other visualisations also do not use the template colours
- in the conclusion, delete the sentence directly under the conclusion heading, is it unnecessary. 
- maybe instead add there how the analysis addressed each of the 3 scenarios. 

Otherwise this is a great first draft!

janzen63122-debug:
Hi @molliefernandez-mentor, thanks for the review! I've addressed your latest points:

- Updated the Use Case Index file.
- Fixed the visualisations 
- Folium map icons now use template colours (green markers).
- Applied template colours to the remaining visualisations that were on defaults.
- Removed the unnecessary sentence under the Conclusion
- Added a section to the conclusion explaining how the analysis addresses each of the three scenarios 

Ready for re-review when you have a chance — thanks!

ashleyyy01 (changes requested):
Great progress @janzen63122-debug thanks for the updates!

I still can't see some of the visualisations, same as Mollie mentioned early on, the ones using fig.show() aren't displaying. Matplotlib's plt.show() charts don't have this problem since they save as static images. I believe the Plotly charts are being saved as interactive HTML rather than an image, but since we're only keeping the .ipynb file, could you save these as static images instead?

A few other things:
- Small mismatch in the content title for section 4 it says "Median house prices by suburb" but the actual heading reads "Median Price of Houses in each Suburb".
- Sections 2.2/2.3 (null handling, outlier trimming) I see you explain what's being done but  maybe include a short line on why it matters for the analysis.
- Average transaction line uses plain red (color='r') instead of a template colour and "Most Common Theme Count for Each Suburb" chart has no colour set, so it'll default to Plotly's standard blue. Maybe you missed these because you cannot directly see them in the ipynb file so just change this to the template colour scheme please. 
- Typos: "traansactions" in (section 3.1)
- Couple of US spellings in the Scenario/Background sections — "analyze"/"analyzing" and "shopping centers" should be "shopping centres".

Keep up the good work, awaiting the new changes!

janzen63122-debug:
Hi @ashleyyy01 thank you for the feedback. I have i have amended all the stuff above. you may take a look. 
Hope can finish this before trimester end!

molliefernandez-mentor (changes requested):
Hi @janzen63122-debug 

**Author**: {Author/s}
**Level**: {Intermediate}
- Delete the background section, the content is nor background. 
- I am getting an error in 3.1
Image

janzen63122-debug:
@molliefernandez-mentor done on the first 2 points. 
For the error 3.1, you can try to install the kaleido package   '!pip install -U kaleido' so the static image can display in the ipynb

ashleyyy01 (changes requested):
Hey @janzen63122-debug, thank you for the changes! You've already got the fix sorted with !pip install -U kaleido, could you just add that directly into the .ipynb file itself, near the top before the imports? Adding that install cell should fix this and every other fig.show() chart at once. We need the notebook to be fully ready to go on its own, without anyone needing to install anything separately. 

Thats the only thing left now I would say, a final look at all the maps and we should be ready to finish this! Thank you!

molliefernandez-mentor (changes requested):
Hi @janzen63122-debug 

- Please update the author section so it is just "**Author**: Yash K Daware and Vaibhavi J Deo"
- The use case needs to stand on its own and run without errors, it is not for me to fix the errors and run it. You need to make sure it is ready to run, it is a tutorial for people new to Data science, we don't want them to be running into errors before they even begin. 
SO as Ashley suggested please make the changes and ensure it will run for anyone :)

janzen63122-debug:
@ashleyyy01 @molliefernandez-mentor hi, i have updated as said. should be the final revised before merge.

molliefernandez-mentor (changes requested):
Hi @janzen63122-debug 

I am still unable to run the code:

Image

VedS8:
Hey Janzen, found the actual cause of why Mollie can't run the notebook, plus a couple of small things while I was checking it.

**Why it won't run for Mollie:** Her traceback shows the error is being raised inside Plotly's `_kaleido.py`, at a check used by older Plotly versions that still expect Kaleido's legacy `Scope` object. Kaleido v1.0 removed that object, so when an older Plotly tries to use it, the scope creation fails and Plotly falls back to the generic "please install kaleido" message, even though Kaleido is technically installed, just the wrong major version for that Plotly version.

The traceback also shows she's on Python 3.13, and the older compatible Kaleido version (`0.2.1`) predates Python 3.13, so pinning Kaleido to that version likely won't work. Could you please upgrade `plotly` to `5.24` or later instead, since that supports the newer Kaleido API? That should allow `!pip install -U kaleido` to keep installing the latest Kaleido without the version conflict.

**Two small typos left:**

1. "Maximizing success and profitability" in Scenario 2 should be "maximising".
2. "Enhance its appeal and livability" in Scenario 3 should be "liveability".

Thank you!

PR #2129, opened 2026-09-04
DS: Usecase Update UC00041

AtishayDeakin:
Renamed to `Urban Flood Management and Emergency Response (2026)`.

The notebook previously ended after loading and inspecting the datasets, so the analysis itself has been built rather than patched.

New analysis (previously missing)

Builds a flood risk score per ~100m grid square from three independent signals:
- Capacity risk — pipe diameter (narrowest quartile = highest risk)
- Condition risk — recorded pipe condition (worst quartile = highest risk)
- Density risk — stormwater pit + drainpipe count per square (sparsest quartile = highest risk)

Signals are combined by averaging only across those with data, rather than zero-filling gaps, with a `sources_count` confidence column retained. Output is an interactive flood risk heatmap plus a ranked list of priority locations.

Reviewer fixes
- Removed the `from config import API_KEY` dependency — the public export endpoint needs no key
- Dropped `city-of-melbourne-3d-textured-mesh-photomesh-2020`, it returns metadata-only columns via CSV export. Reasoning documented in Section 2 rather than silently removed
- Removed the duplicate `pandas` import and the unused `squarify` import
- Converted US spellings to Australian English

Template alignment
- Rebuilt against the current template: header bar, metadata, Scenario / What this teaches / Background / Data Sets / Contents, 8 numbered sections with description lines, Conclusion
- Duration, level and pre-requisite skills stated; datasets listed with links

Note on 8.2

Ranking purely by score put the least-evidenced squares at the top — squares with stormwater pits but no recorded drainpipes score high on density alone. 8.2 now reports high-confidence locations (all three signals) separately from the 277 no-drainpipe squares, which are flagged as a data gap rather than confirmed risk.

Known assumption: the `condition` field's direction (higher = worse) follows common council convention but isn't documented in the dataset. Flagged in Section 5.2 and the Conclusion; worth verifying against Melbourne's asset management documentation.

Testing: Full Restart & Run All — 20 cells, sequential, no errors or warnings.


s224251967-hub (changes requested):
Hi Atishay, 
Thanks for your update. Below is feedback.

- Ensure the date in the filename reflects the analysis data range where that range is known. 
- Keep “Date Created” as the notebook creation date. Do not use the current trimester.
- Remove the detailed discussion about dropping the 3D textured mesh photomesh dataset. It is not needed for the final tutorial.
- Remove update-history wording such as “the earlier from config import API_KEY import has been removed”. Explain the current code and data-access approach.
- Fix the heatmap score scale and colours. Section 8.1 passes a flood risk score of approximately 1–4 to HeatMap, while the gradient is defined from 0.0 to 1.0. Scores above 1 may all display using the highest colour. Normalise the score to 0–1 before creating the heatmap.

molliefernandez-mentor (commented):
Hi @AtishayDeakin 

- Remove the brackets around 2026 in the title
- have you put 2026 in the title because it is the current year or because the analysis is for the current year. Because it seems like the datasets were most recently updated in 2023 and are obviously not live datasets. 
- Update the author and date created as discussed in teams
- In the scenario section, before the 2 questions can you please put like a "This use case aims to answer the following questions:"
- The sub headings in the conclusion should just be bolded body text not sub headings
- in the conclusion section can you please address how you have answered the 2 questions posed in the scenario
- Please move your use case into 2nd position int he use case index table so that they are in order of use case number and add the author. 
- Please rebase your branch.

Otherwise it's looking great!

AtishayDeakin (commented):
Hi @molliefernandez-mentor @s224251967-hub,

I have done the requested changes, please have a look at it.

s224251967-hub (changes requested):
Hi @AtishayDeakin,
Please remove the remaining reference to the 3D mesh dataset from the Conclusion, as it is no longer used; otherwise, everything looks good to me.

AtishayDeakin (commented):
Hi @s224251967-hub, 

I have removed the remaining reference to the 3D mesh dataset from the Conclusion.

molliefernandez-mentor (approved):
Thanks @AtishayDeakin looks good to me, once you get @s224251967-hub's final approval, you can merge!

s224251967-hub (approved):
Thanks @AtishayDeakin, Happy to see it's ready to merge.

PR #2131, opened 2026-09-06
WD: Sprint 3 Task 8 – Next.js Asset Optimization

HimanshuPyakurel:
This PR improves frontend asset handling and performance by updating image and font usage across the Next.js application.

Changes completed:
- Replaced active raw HTML `` tags with the Next.js `<Image />` component.
- Added appropriate image sizing and handling for dynamic/preview images.
- Updated font loading to use Next.js `next/font`.
- Configured Poppins globally through the layout.
- Configured Barlow Condensed using a shared font variable.
- Removed old direct Google Fonts loading.
- Resolved merge conflicts with the latest master branch.
- Added the required `sonner` dependency from the updated master code.

Verification:
- No active raw `` tags remain.
- Production build completes successfully.
- Desktop and mobile Lighthouse testing completed.
- Branch is up to date and able to merge.


hamizhassan16 (changes requested):
The font migration to next/font is a solid, real win — good removal of the external Google Fonts request.

On the image side, though: most of the  → <Image> conversions add `unoptimized`, which turns off Next's actual optimization (resizing, WebP/AVIF, lazy loading) — the main thing this PR is meant to deliver. I checked next.config.mjs and it already whitelists *.supabase.co/storage/v1/object/public/** via remotePatterns, so for images actually served from there — especially ImageHoverPreview.tsx (used across every admin table) and the gallery grid/lightbox — unoptimized isn't needed and just cancels out the benefit for the most-used component in this PR.

unoptimized does make sense elsewhere (blob preview URLs before upload, base64 data URIs in NotebookRenderer.tsx) — just asking to drop it where the remote host is already configured, so the Lighthouse gains actually apply to the admin panel.

suba-t24 (approved):
Looks great! All requested changes regarding the `unoptimized` prop on Next.js images have been perfectly addressed. Thank you!

hamizhassan16 (approved):
Confirmed the last item is resolved — the profile/page.jsx diff is fully reverted, diffed it against master myself and it's a clean zero. All three things I flagged (Escape key, BlogsTable typing, unrelated file) are addressed. Approving.

PR #2136, opened 2026-09-07
WD: Fix created_by in categories/route.ts

qweasdzxc03:
- Mirrored usecases/shared.js resolveCreatedBy

### DIFF
--- a/next_webapp/src/app/api/categories/route.ts
+++ b/next_webapp/src/app/api/categories/route.ts
@@ -9,6 +9,9 @@ import { errorResponse } from "@/app/api/library/errorResponse";
 import { getAuthUser } from "@/app/api/library/auth";
 import { NextRequest } from "next/server";
 import logger from "@/utils/logger";
+import mongoose from "mongoose";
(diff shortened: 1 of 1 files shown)


suba-t24 (changes requested):
This PR takes a good approach to fixing the `created_by` mapping by looking up the legacy user ID. However, there are a few critical issues that will cause `POST /api/categories` to crash at runtime:

1. **Missing Imports:** `mongoose` and `User` are used in the new code block, but they are not imported anywhere in the file.
2. **Missing Database Connection:** Mongoose queries in Next.js API routes require an active database connection. You need to call your `dbConnect()` utility before making the queries.

Please add the necessary imports and the connection call:

```typescript
import mongoose from "mongoose";
import User from "@/models/mongoose/User"; // Update path if necessary
import dbConnect from "@/lib/dbConnect"; // Update path if necessary

// Inside the POST function, before querying User:
await dbConnect();
```

Once those are added, this should work perfectly.

suba-t24 (approved):
Comments addressed and Looks good

PR #2138, opened 2026-09-08
WD: contributor section fix

suba-t24:
- This PR backfills a large batch of historical contributor data (from 2020 to 2026) and introduces backend schema updates and frontend UI enhancements to better organize and display complex team structures.

Screenshot 2026-09-08 at 1 21 44 pm

- Add pre-validate hook logic to reject position/team


bettycheng14 (changes requested):
Thanks @suba-t24, the sorting and team-role validation logic is solid and the contributors rendered in /about page is great! A couple of things worth sorting out before merge though:
## Changes required: Admin UI can't manage the new feature

`ContributorForm.tsx` in frontend hardcodes its own copy of contributor types/teams/roles, separate from Contributor.ts. This PR adds project_lead, new roles, and Project Team to the model only, but the form wasn't updated, so there's no dropdown option for any of them. These can currently only be created via the API/DB, not the admin panel.

→ Suggest moving TEAM_ROLES, TEAMS, ROLES, and CONTRIBUTOR_TYPES into exisitng `types/contributor.ts` so both files can import from one source of truth. This avoids duplication and keeps Mongoose out of the client-side form.

## Needs clarification: Two definitions of "Project Lead"

The PR adds both contributor_type: "project_lead" (trimester-level) and position: "Project Lead" (a student role under Data Science Team). They behave differently, since only contributor_type drives the "Project Leads:" section.

Could you confirm if both are intentional, or should they be consolidated into one?

suba-t24:
Thanks for the feedback! I've pushed a commit addressing both of your points:

1. Centralized Types: Moved TEAM_ROLES, TEAMS, ROLES, LEVELS, and CONTRIBUTOR_TYPES into src/types/contributor.ts to act as the single source of truth. Both Contributor.ts (Mongoose Model) and ContributorForm.tsx (Frontend Form) now import their types and dropdown options from here, eliminating the duplication.
2. "Project Lead" Clarification: I can confirm "Project Lead" should exclusively be a trimester-level contributor_type. I've removed it from the student positions under the Data Science Team to resolve the conflicting definitions.

I also went ahead and added "Project Leads" to the types filter dropdown on the main /admin/contributors list page so they can be easily filtered alongside Students and Mentors.

Let me know if there's anything else needed!

hamizhassan16 (changes requested):
The move to a single source of truth in types/contributor.ts is a good cleanup, but a few things need fixing before merge:

1. src/lib/getContributors.ts line 2 still imports { TeamName } from "@/models/mongoose/Contributor", but this PR removes that export from the model. Import TeamName from "@/types/contributor" instead (you already import other types from there on line 4). It's a tsc error today, only hidden because ignoreBuildErrors is on.

2. ContributorForm.tsx line 246: TEAM_ROLES[formData.team] is now an implicit-any index error since TEAM_ROLES became `as const` (it used to be Record<string, string[]>). Cast the key to TeamName or add a typed lookup helper.

3. The new pre("validate") position/team check (and the existing enum on the position field) will reject any student whose role isn't in the current TEAM_ROLES lists. How does the 2020-2026 backfill mentioned in the description get inserted - raw Mongo insert, or does the historical data already conform to the new vocabulary? Want to make sure the backfill doesn't silently fail on legacy role names.

4. ContributorsSection.tsx line 231: "Project Lead(s):" is a hardcoded English string while the rest of that card uses t("contributors.…"). Please add a translation key for it.

5. Minor: the mentors <p> margin was changed from mb-4 to mb-2 unconditionally - when there are mentors but no project leads, the spacing before the team list shrinks. Cleaner as a conditional margin.

Also, the PR description mentions a large historical data backfill but the diff is code-only - is the data import happening separately / directly against the database?

suba-t24:
Thanks for the great feedback! I've addressed all the points and pushed the fixes:

1. Centralized Types: Updated src/lib/getContributors.ts to import TeamName directly from src/types/contributor.ts instead of the Mongoose model, fixing the hidden tsc error.
2. Form Type Error: Added an explicit cast (formData.team as TeamName) in ContributorForm.tsx to fix the implicit-any index error on TEAM_ROLES.
3. Data Backfill & Validation: Good catch regarding the backfill! The historical data import was done directly against the MongoDB database via scripts (raw Mongo inserts). Because of this, it bypasses the Mongoose application-level pre("validate") hooks, so it won't fail on legacy role names. The strict schema rules will just ensure that any new entries or edits made moving forward conform to the new standard vocabulary.
4. Translations: Added the missing translation keys (projectLead and projectLeadsPlural) to all language files in messages/*.json and wired them up in ContributorsSection.tsx.
5. Mentor Spacing: Updated the mentors section to use a conditional margin. It now applies mb-2 if there are Project Leads below it, and mb-4 if there aren't, which keeps the spacing clean!

Let me know if anything else is needed before we merge!

hamizhassan16 (changes requested):
Good progress - the TeamName import, the TEAM_ROLES index cast, and the i18n wiring for "Project Lead(s)" are all fixed properly (confirmed the component now goes through t("contributors.projectLead"/"projectLeadsPlural") instead of a hardcoded string).

One thing left: the two new translation keys were added to all 8 locale files, but only the English file has real translated text - ar.json, es.json, and the rest just have the literal English strings ("Project Lead: {names}") copy-pasted in, unlike mentoredBy which has proper translations in each. Worth getting real translations in before merge, or flagging it as a known follow-up if that's handled separately.

Also still an open question from before: how does the 2020-2026 historical backfill interact with the new team/position validation hook in Contributor.ts? That file is unchanged in this commit, so I don't think it's been answered yet.

suba-t24:
Thanks for the review! I've just pushed a commit that addresses both of the remaining items:

Translations: Added the actual translations for projectLead and projectLeadsPlural across all 7 non-English locale files, so we don't have to follow up on this later.

Historical Backfill Validation: I updated the validation hook in Contributor.ts to skip the strict team/position enum checks if an is_legacy_backfill flag is set. This ensures that standard application data remains strictly validated at the database level, while allowing the 2020-2026 legacy backfill script to pass seamlessly by just including is_legacy_backfill: true with those records.

hamizhassan16 (approved):
All outstanding points are resolved:

1. Translations are real now, not placeholders - spot-checked Arabic, Chinese, Spanish, and the rest, all have genuine translated text for the new project-lead keys.

2. The historical backfill question has a proper answer: the new is_legacy_backfill flag skips strict validation for grandfathered records. Checked both POST /api/contributors and PUT /api/contributors/[id] - they build their Contributor.create()/.set() payloads from an explicit field allowlist, and is_legacy_backfill isn't in either one, so it can't be set or cleared through the app's API at all - only a script writing directly via the model could touch it. No validation-bypass exposure to admins or the public API.

3. The mentors-paragraph margin is conditional now instead of unconditionally shrunk.

Approving.

hamizhassan16:
@bettycheng14  kindly review as soon as possible

bettycheng14 (approved):
Thanks @suba-t24,
Both points addressed: types centralized into types/contributor.ts and project_lead correctly consolidated as a contributor_type only. 

But there will be some failing Cypress tests regarding admin workflows need to be fixed after this pr merge. I'll patch those separately.

Good work, happy to merge!

PR #2139, opened 2026-09-08
WD: security: harden forgot-password flow

suba-t24:
Replace Math.random() with crypto.randomInt() for temp password generation
- Add reset_token_expires (15 min TTL) and reset_token_used flag to User model
- Add per-email and per-IP rate limiting (3 requests / 15 min window)
- Fix nodemailer transport: add secure:false for correct STARTTLS on port 587
- Ensure DB token and emailed token are always generated atomically in the same request

Verified and tested locally for the account: tsubi1024@gmail.com

### DIFF
--- a/next_webapp/src/app/api/auth/forgot-password/route.ts
+++ b/next_webapp/src/app/api/auth/forgot-password/route.ts
@@ -4,16 +4,43 @@ import User from '@/models/mongoose/User';
 import bcrypt from 'bcryptjs';
 import nodemailer from 'nodemailer';
 import { errorResponse } from '@/app/api/library/errorResponse';
+import crypto from 'crypto';
 
 const EMAIL_REGEX = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
 const TEMP_PASSWORD_CHARS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
 const TEMP_PASSWORD_LENGTH = 10;
 
+const rateLimitCache = new Map<string, { count: number, resetTime: number }>();
+const RATE_LIMIT_MAX_REQUESTS = 3;
+const RATE_LIMIT_WINDOW_MS = 15 * 60 * 1000; // 15 minutes
+
+function checkRateLimit(key: string): boolean {
+    const now = Date.now();
+    const record = rateLimitCache.get(key);
+    
+    if (!record) {
+        rateLimitCache.set(key, { count: 1, resetTime: now + RATE_LIMIT_WINDOW_MS });
+        return true;
+    }
+    
+    if (now > record.resetTime) {
+        rateLimitCache.set(key, { count: 1, resetTime: now + RATE_LIMIT_WINDOW_MS });
+        return true;
+    }
+    
+    if (record.count >= RATE_LIMIT_MAX_REQUESTS) {
+        return false;
+    }
+    
+    record.count++;
+    return true;
+}
+
 function generateTempPassword(): string {
     let result = '';
     for (let i = 0; i < TEMP_PASSWORD_LENGTH; i++) {
         result += TEMP_PASSWORD_CHARS.charAt(
-            Math.floor(Math.random() * TEMP_PASSWORD_CHARS.length),
+            crypto.randomInt(0, TEMP_PASSWORD_CHARS.length)
         );
     }
     return result;
@@ -26,6 +53,11 @@ const SAFE_RESPONSE = NextResponse.json(
 
 export async function POST(request: Request) {
     try {
+        const ip = request.headers.get('x-forwarded-for') || request.headers.get('x-real-ip') || 'unknown';
+        if (ip !== 'unknown' && !checkRateLimit(`ip_${ip}`)) {
+            return errorResponse('Too many requests, please try again later', 429, 'RATE_LIMIT_EXCEEDED');
+        }
+
         const { email } = await request.json();
 
         // 1. Validate input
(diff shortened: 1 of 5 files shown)


bettycheng14:
Thanks for the changes @suba-t24! I went through the new reset-token flow. There are **2 changes required before merging**, plus a few follow-ups.

## Changes required

### 1. Temporary password logged on SMTP failure

In `next_webapp/src/app/api/auth/forgot-password/route.ts`, `tempPassword` is logged in the SMTP error handler.

Could we avoid logging `tempPassword` in production, and only include it for development/local testing if needed? Otherwise, an SMTP failure could expose the plaintext temporary password in production logs.

### 2. `isSameAsTemp` uses the wrong field

In `next_webapp/src/app/api/auth/reset-password/route.ts:97`, `isSameAsTemp` compares `new_password` with `userData.password`.

Since the temporary password hash is now stored in `userData.reset_token`, this should compare against `userData.reset_token` instead.

With the current logic, users can reuse the temporary password as their new password, while their previous password could incorrectly trigger the "same as temp password" check.

---

## Follow-ups

**3. Rate limiting on `reset-password`**

`forgot-password` has the 3 requests / 15 min limit, but `reset-password` doesn't. Since this is where the temporary password is checked, some throttling here could help prevent repeated guessing attempts.

**4. Automated tests**

The verification was done manually. It would be good to add tests for the `reset-password` flow, especially the temporary-password comparison, which would have caught issue 2.

Overall, the new rate limiting and token expiry are good improvements. Once the 2 required changes are addressed, I think the rest can be followed up separately.

hamizhassan16 (changes requested):
The core fix here is genuinely important — previously any forgot-password request immediately overwrote the real password, which is a trivial account-lockout DoS. Separating out reset_token with expiry + single-use, and switching to crypto.randomInt for the temp password, are both real, correct fixes.

Two things I'd like addressed before merging:
1. The rate limiter uses an in-memory Map, which won't hold state across server instances or survive a serverless cold start. #2141 (login rate-limiting) solves the same class of problem with a Mongo-backed, TTL-indexed store — worth using the same pattern here instead of a second, weaker implementation.
2. On SMTP failure, the temp password gets logged to the console "for dev testing." For a hardening PR, that's a plaintext credential leak risk in production if email ever fails — can this be gated behind NODE_ENV !== 'production' or removed?

Requesting changes on those two, the token-separation fix itself is solid.

suba-t24:
Thanks for the feedback, I have addressed them:
1. Robust Rate Limiting: Replaced the in-memory rate limit Map with a new Mongoose model (PasswordResetRateLimit.ts) utilizing a TTL sliding window index, following the same robust pattern set in PR #2141. It tracks both the number of forgot_password_requests per 15-minute window (limit 3 per IP/Email) and failed_reset_attempts (limit 5). This ensures our rate limits will persist across Cloud Run instances and cold starts!
2. Fixed Password Leak: The dev fallback console.error that outputs the generated plaintext password during an SMTP failure in the forgot-password flow is now strictly gated by if (process.env.NODE_ENV !== 'production'), eliminating the risk of credential leaks in prod logs.
3. Fixed isSameAsTemp Logic: Fixed a bug where reset-password was comparing the new_password against the permanent password instead of the temporary reset_token.

hamizhassan16 (changes requested):
Both earlier points are addressed - the rate limiter is now Mongo-backed with a TTL sliding window (matching the #2141 pattern), and the temp-password console log is gated behind NODE_ENV !== 'production'. Token separation core is solid.

One issue before merge: recordPasswordResetAttempt does an unconditional $inc: { attempts: 1 } on the upsert. If a key already hit its limit, then goes quiet past the 15-minute window, but Mongo's TTL sweep hasn't physically removed the row yet (sweep runs ~every 60s and lags under load), checkPasswordResetRateLimit correctly ignores the stale row via isWithinWindow - but the record step then increments it (e.g. 3 -> 4) and stamps last_attempt_at = now, so the next request locks the user out after a single real attempt. Same stale-counter resurrection issue raised on #2141. Fix: when recording, if the existing row's last_attempt_at is outside the window, set attempts: 1 instead of incrementing.

Minor: the explicit await dbConnect() in both routes is now redundant (the rate-limit helpers connect themselves); and the forgot_password_request counter is only cleared by TTL, never on success - assuming that's intentional?

suba-t24:
Updates pushed addressing all feedback:

1. Fixed TTL Race Condition: Replaced the blind $inc with a Mongo aggregation pipeline that resets attempts to 1 if the existing row is older than 15 mins.
2. Fixed Sliding Lockout: Moved the attempt recording to execute before the 429 return so the lockout window correctly slides forward under attack.
3. Fixed IP Spoofing: getClientIp now grabs the last IP from the chain instead of the first, preventing trivial bypassing via fake headers.
4. Removed the redundant dbConnect calls in both routes.
5. Added logic to clear the forgot_password_request rate limit upon successful email dispatch so valid users aren't punished.

hamizhassan16 (approved):
The stale-counter issue is fixed - recordPasswordResetAttempt now uses a conditional aggregation-pipeline update that resets attempts to 1 when the existing row's last_attempt_at is outside the window, instead of blindly incrementing. Checked the within-window, stale-doc, and upsert-insert paths and all three produce correct values. Redundant dbConnect calls, clear-on-success, and IP parsing are all handled too.

Approving. Two non-blocking notes:
1. getClientIp now uses the last x-forwarded-for entry instead of the first - better for anti-abuse, but double-check it matches your Cloud Run / LB topology (on some setups the real client IP is second-to-last). Also worth aligning with #2141's getClientIp, which still uses the first entry.
2. The insert path relies on a missing last_attempt_at making the $lt condition true - it works, just noting it's implicit rather than an explicit insert branch.

suba-t24:
Thanks for the notes! Just pushed a quick follow-up to address both:
- Reverted `getClientIp` to strictly use the first entry. This perfectly aligns with #2141 - will check with Lakindu, and raise a followup PR if decided otherwise
- Added an explicit `$ifNull` check to the `$lt` aggregation pipeline so missing documents now explicitly default to Epoch 0 rather than implicitly relying on BSON's comparison quirks.

PR #2141, opened 2026-09-08
WD: Add rate-limiting to the login endpoint

lakindubd:
What this does

Adds rate-limiting to the login endpoint (/api/auth/login), which previously had no throttling at all an attacker could hammer login as fast as the network allowed, for any account, indefinitely. Now repeated failed attempts are tracked and locked out.

How it works

- **Per-email:** 5 failed attempts within 15 minutes to locked for 15 minutes. Stops an attacker brute-forcing one specific account.
- **Per-IP:** 20 failed attempts within 15 minutes to locked. Stops credential-stuffing (many emails from one source). Looser than the per-email limit so a shared network doesn't lock out legitimate users.
- On lockout, returns **429 Too Many Requests** with a generic message (doesn't reveal whether the email or IP key tripped).
- A successful login **resets** both counters.
- Failed attempts increment on both an invalid email *and* an invalid password (both still return the same 401 INVALID_CREDENTIALS, so the lockout doesn't leak which side failed).

Design decisions

- **Mongo-backed, not in-memory.** The app deploys to Cloud Run (stateless, multiple instances, scales to zero) an in-memory counter would give each instance its own quota and reset on every cold start, so it'd be useless in production. Attempt state lives in a new login_attempts collection so the limit holds across all instances.
- **Sliding window via TTL index**  each failed attempt pushes the document's 15-minute expiry forward, so an active lockout extends as long as attempts keep coming, and the record auto-expires once attempts stop. A read-time recency check guards against MongoDB's TTL sweep lag.
- **Fail-open**  if the rate-limit store itself errors, the login attempt is allowed through rather than blocking all logins. 
- **IP from x-forwarded-for/x-real-ip**  same headers the existing middleware reads.

Files

- **New** src/models/mongoose/LoginAttempt.ts the attempt-tracking model (compound unique key, TTL index).
- **New** src/app/api/library/loginRateLimit.ts  check / increment-on-failure / reset-on-success helpers, thresholds as named constants, all DB access wrapped fail-open.
- **Modified** src/app/api/auth/login/route.ts checks the limit before auth, increments on failure, resets on success. Existing auth flow (user lookup, bcrypt, JWT) unchanged.

## Testing

Verified end-to-end against Atlas:
- 5 failed password attempts on one email to 6th returns 429
- Successful login clears the counter (attempt doc removed from login_attempts)
- login_attempts collection populates with per-email and per-IP docs
- Normal successful login unaffected

### DIFF
--- a/next_webapp/src/app/api/auth/login/route.ts
+++ b/next_webapp/src/app/api/auth/login/route.ts
@@ -2,6 +2,12 @@ import { NextResponse } from 'next/server';
 import bcrypt from 'bcryptjs';
 import jwt from 'jsonwebtoken';
 import { errorResponse } from '@/app/api/library/errorResponse';
+import {
+    checkLoginRateLimit,
+    getClientIp,
+    recordFailedLoginAttempt,
+    resetLoginAttempts,
+} from '@/app/api/library/loginRateLimit';
 import dbConnect from '@/lib/dbConnect';
 import User from '@/models/mongoose/User';
 
@@ -19,10 +25,19 @@ export async function POST(request: Request) {
             return errorResponse('Email and password are required', 400, 'MISSING_FIELDS');
         }
         const normalizeEmail = email.toLowerCase().trim();
+        const ip = getClientIp(request);
+
+        // 1.5. Rate limit checked before any lookup/compare below.
+        const { limited } = await checkLoginRateLimit(normalizeEmail, ip);
+        if (limited) {
+            return errorResponse('Too many attempts, please try again later.', 429, 'TOO_MANY_ATTEMPTS');
+        }
+
         // 2. Find user by email
         const userData = await User.findOne({ email:normalizeEmail }).exec();
 
         if (!userData) {
+            await recordFailedLoginAttempt(normalizeEmail, ip);
             return errorResponse('Invalid email or password', 401, 'INVALID_CREDENTIALS');
         }
 
@@ -33,6 +48,7 @@ export async function POST(request: Request) {
         );
 
         if (!isPasswordValid) {
+            await recordFailedLoginAttempt(normalizeEmail, ip);
             return errorResponse('Invalid email or password', 401, 'INVALID_CREDENTIALS');
         }
 
@@ -65,6 +81,9 @@ export async function POST(request: Request) {
         };
 
         const token = jwt.sign(tokenPayload, JWT_SECRET, { expiresIn: '7d' });
+
+        // Successful login clear any tracked failed attempts.
+        await resetLoginAttempts(normalizeEmail, ip);
 
         // 7. Return success response with everything
         return NextResponse.json(
--- a/next_webapp/src/app/api/library/loginRateLimit.ts
+++ b/next_webapp/src/app/api/library/loginRateLimit.ts
@@ -1,0 +1,128 @@
+import dbConnect from "@/lib/dbConnect";
+import LoginAttempt, {
+  LOGIN_ATTEMPT_WINDOW_SECONDS,
+} from "@/models/mongoose/LoginAttempt";
+import logger from "@/utils/logger";
+
+// Failed-attempt thresholds per key type. Both share the same 15-minute
+// window as the LoginAttempt TTL (src/models/mongoose/LoginAttempt.ts).
+const EMAIL_MAX_ATTEMPTS = 5;
+const IP_MAX_ATTEMPTS = 20;
+const WINDOW_MS = LOGIN_ATTEMPT_WINDOW_SECONDS * 1000;
+
+export interface RateLimitCheck {
+  limited: boolean;
+}
+
+function isWithinWindow(lastAttemptAt: Date): boolean {
+  return Date.now() - lastAttemptAt.getTime() < WINDOW_MS;
+}
+
+/**
+ * Extract the client IP the same way src/middleware.ts does: first entry of
+ * x-forwarded-for, else x-real-ip. `request.ip` is not reliable on Cloud
+ * Run TLS terminates at a proxy in front of the instance.
+ */
+export function getClientIp(request: Request): string {
+  const forwardedFor = request.headers.get("x-forwarded-for");
+  if (forwardedFor) {
+    const first = forwardedFor.split(",")[0]?.trim();
+    if (first) return first;
+  }
+  return request.headers.get("x-real-ip") || "unknown";
+}
+
+/**
+ * Check whether the given (already-normalized) email or IP is currently
+ * locked out. Fails open: any store error is logged and treated as
+ * "not limited" so a Mongo outage can never block logins outright.
+ */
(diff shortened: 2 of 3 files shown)


hamizhassan16 (changes requested):
Nicely done, the Mongo-backed store with the sliding-window TTL is the right way to do this (and a much better pattern than the in-memory approach in #2139, worth pointing that PR at this one). Fails open on a DB outage is the correct tradeoff too.

One small thing: this also deletes api/chat/route.ts, an unrelated dead stub. Harmless, but out of scope for a rate-limiting PR, mind adding a line on why it's in here?

Can you please make changes before we merge to keep repo clean, will be happy to approve once changes are made

lakindubd:
@hamizhassan16 this branch was based on the chat-deletion branch, which has since been reviewed and merged separately (#2140, Suba approved it). So the deletion was already handled on its own it was just showing here because of the shared branch base. vie rebased onto master to drop it. the deletion is out of the pr now, its purely rate-limiting changes.

suba-t24:
Solid architectural choices with MongoDB and the fail-open design! However, there are three critical logic bugs in how the limits are enforced that need to be fixed before merging:

1. TTL Sweeper Race Condition: recordFailedLoginAttempt blindly does an $inc: { attempts: 1 }. Because Mongo's TTL sweep runs every ~60s and lags under load, a request right after the 15-minute mark can increment a stale counter instead of starting fresh, causing an instant premature lockout. (Fix: use a $cond aggregation pipeline to set attempts to 1 if last_attempt_at is older than the window).
2. Lockout Doesn't Slide: In route.ts, returning 429 early means recordFailedLoginAttempt is never called during an active lockout. The last_attempt_at timestamp never bumps, so the lockout strictly expires 15 mins after the 5th attempt rather than extending under attack.
3. IP Spoofing Bypass: In getClientIp, taking .split(",")[0] on x-forwarded-for allows attackers to bypass the per-IP limit entirely just by sending a fake IP header, since Cloud Run appends the real IP to the end of the chain. It needs to grab the last IP instead.

I just pushed fixes for these exact same three issues in my forgot-password rate-limiter branch if you want to use that as a reference for the fixes here!

hamizhassan16:
Thanks for dropping the api/chat/route.ts deletion - that part's resolved.

But the main issue is still open: recordFailedLoginAttempt still does an unconditional $inc: { attempts: 1 }. This is the stale-counter resurrection bug Copilot flagged (suppressed comment on loginRateLimit.ts:83) - a doc past the 15-min window that hasn't been TTL-swept yet gets incremented instead of reset, so checkLoginRateLimit ignores it via isWithinWindow but the next request sees attempts >= max with a fresh last_attempt_at and locks the user out after a single real attempt.

#2139's recordPasswordResetAttempt was just reworked to fix the identical issue - it uses a conditional aggregation-pipeline update that sets attempts to 1 when last_attempt_at is outside the window, else increments. Please apply the same pattern here so the two PRs are consistent.

Also minor: getClientIp uses the first x-forwarded-for entry here, but #2139 now uses the last - worth aligning. And Copilot's second point (sliding window not extended while already rate-limited) is still worth a quick acknowledgement even if you decide it's out of scope.

lakindubd:
image
image
recordFailedLoginAttempt now uses a conditional aggregation-pipeline update
locked-out requests now bump last_attempt_at before returning 429, so continued attempts actually extend the lockout
getClientIp now takes the last x-forwarded-for entry
Ready for re-review whenever you get a chance. @hamizhassan16

champa1n73 (commented):
Nice work, the DB-backed sliding-window limit and reset-on-success approach make sense, and the Atlas tests are reassuring. Before merging, there are a few important points to address: please consider the risk of attackers creating many unique email keys and flooding the database with temporary TTL records; make sure `X-Forwarded-For`/`X-Real-IP` are only trusted when requests come through a trusted proxy; ensure the limit increment and threshold check are handled atomically to avoid concurrent requests bypassing the limit; and add high-priority logging or alerts for cases where the rate-limit store fails and the system falls back to allowing requests. Addressing these should improve the security and reliability of the implementation.

lakindubd:
image
image

Pushed the fixes for all four review points. Test evidence for the key one (the concurrent-bypass fix) above:

- Fired 10 simultaneous failed-login requests at the same email from a clean count.
- Result: exactly 5× 401 (allowed through, failed auth) and 5× 429 (rejected as over-limit).
- Before the fix, all 10 would have passed the check and returned 401, the atomic increment-then-check now enforces the cap under concurrent load, not just serialised requests.

Summary of what changed:
- **Atomicity**  limit decision now made from the atomic findOneAndUpdate's returned count, not a prior read (tested above).
- **Flooding**  a rate-limited request now only records the key that triggered the limit, so an over-limit IP rotating novel emails no longer creates a doc per attempt.
- **Fail-open**  added a distinct RATE_LIMIT_FAIL_OPEN code + Sentry capture so a store outage silently disabling the limiter is now alertable.
- **Trusted proxy**  documented the fully-managed Cloud Run assumption (last XFF entry is the trusted one), noted to revisit if a CDN/proxy is added.
- Also raised the per-IP limit 20→100 so a shared-NAT burst of legitimate logins doesn't get spuriously locked out.

@champa1n73 please have a look when possible thanks!

suba-t24 (approved):
Everything looks perfectly secure, atomic, and extremely well-tested. This PR is in excellent shape!

HarshanaDeakinUni (commented):
LGTM,  the atomic findOneAndUpdate for the check-and-increment holds up under concurrent load. tests make sense forthis, the sliding TTL window is the right call for Cloud Run's stateless instances, and the fail-open + Sentry alerting means a Mongo thing degrades gracefully instead of locking everyone out which is perfect. Good to merge to me.

HarshanaDeakinUni (approved):

champa1n73 (approved):
The major issues previously identified in review appear to have been fixed, and the latest implementation is substantially better. Good work!

hamizhassan16 (approved):
The latest commit is a real, well-executed hardening pass. It closes a genuine concurrency bug: the previous version read the attempt count, then separately incremented it later on failure, leaving a window where concurrent requests could race past the limit. The rewrite makes checkLoginRateLimit itself an atomic findOneAndUpdate increment (upsert: true, new: true), so there's no read-then-write gap to race.

Also checked: IP is checked/incremented before email so an IP-limited attacker rotating through novel emails doesn't create a fresh doc per request; the IP ceiling was raised to 100 with a clear rationale for shared-NAT/lab false positives; Sentry alerting was added on fail-open so a DB outage silently disabling rate limiting actually gets noticed; and the x-forwarded-for trust assumption (last entry, safe specifically because of the Cloud Run managed topology) is now documented in a comment. Traced the threshold math and confirmed the effective lockout point is unchanged - still exactly 5 genuine attempts before block. No dangling references to the removed recordFailedLoginAttempt.

Approving.

PR #2142, opened 2026-09-08
WD: Consolidate Playground and Archive with T#_YYYY convention and dedupe against usecases

hamizhassan16:
Renamed 40 outdated Playground/ folders to the team's T#_YYYY_<name> convention (determined from git commit history, cross-checked against folders whose names already stated their trimester) and moved them into usecases/UPDATE NEEDED/.

Before moving, checked every file's content hash against the rest of the repo and removed only exact duplicates of already-published work:
- 27 files duplicating content already in usecases/READY TO PUBLISH or FINALISED
- 1 dataset duplicating usecases/DEPENDENCIES
- 110 files where an entire project (Bridge and Road Crack Detection) had already been copied into usecases/UPDATE NEEDED elsewhere

For that Crack Detection project, migrated over what the other copy was actually missing - a 163MB training dataset and a self-contained demo pipeline - before removing the rest.

Active team members' folders (the 12 marked "In Team" in the PR #2054 cleanup tracker) were left untouched, staying in Playground under their original names. One folder covering two unrelated projects (Saneli Wijesinghe) was split into two, each named for its actual project.

Extended the same process to Playground/ARCHIVE/: cross-referenced its 10 folders against the team's Use Case Quality Tracker (244 tracked use cases). 5 matched a tracked use case - 3 were already published, so those folders were removed; the other 2 (UC00021, UC00074) were renamed with their use case code and moved into UPDATE NEEDED with a trimester/year prefix. The remaining 7 folders didn't match anything tracked - renamed with an _unknown_project suffix and moved into UPDATE NEEDED for someone to properly triage later. ARCHIVE/ is now empty.

Playground/ now contains only the 12 active team members' folders - everything else was either deduplicated or relocated.


suba-t24 (approved):
This is a phenomenal cleanup effort! 
The sheer number of files in the diff looks terrifying at first, but digging in, it's clear why: cleaning up those accidentally committed `node_modules` alone wiped out almost 2.5 million lines of bloat from the repository. The rest of the diff correctly tracks the folder renames to enforce the `T#_YYYY_<name>` convention without touching any core application code.
Thank you for taking the time to deduplicate all this and standardize the folder structure. Huge win for the health of the repo. Approving!

molliefernandez-mentor (approved):
Thanks Hamiz, great work!

PR #2143, opened 2026-09-08
WD: optimized third-party scripts with next/script

Muhib1-lewis:
- Theme-init script now uses strategy="beforeInteractive" (prevents flash 
  of wrong theme)
- Instagram embed now uses strategy="lazyOnload" (defers loading until 
  page is idle)
- Audited the app for other third-party scripts i found none found beyond 
  these two

### DIFF
--- a/next_webapp/src/app/[locale]/layout.tsx
+++ b/next_webapp/src/app/[locale]/layout.tsx
@@ -2,6 +2,7 @@ import "./globals.css";
 
 import { NextIntlClientProvider } from "next-intl";
 import { getMessages } from "next-intl/server";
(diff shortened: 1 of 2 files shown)


hamizhassan16 (approved):
Verified both changes properly — beforeInteractive + the required id on the theme-init script is correct, and I checked scriptLoadDisabled is a real documented prop on InstagramEmbed and instgrm.Embeds.process() is the right way to manually hydrate after lazy-loading the script. Also grepped the app for other third-party scripts to sanity-check the audit claim — didn't find anything missed. Approving.

PR #2134, opened 2026-09-07
WD: Sprint Frontend Task 7 + BundleAnalyzer

joshNewn01:
- installed and configured bundleAnalyzer
	 - off by default, to run the analyzer you can toggle it on with the PowerShell command '$env:ANALYZER="true"; npm run *' and then toggle it off by entering the same command in your next call, but with the Boolean set to "false"
 - removed unused import of HiSun/HiMoon and unused toggleTheme() function from ./[locale]/privacypolicy/page.tsx
 - shifted loading html2canvas and jspdf to be within the downloadPDF function in ./[locale]/privacypolicy/page.tsx to reduce initial load by loading only when required
 - created the new file ./components/StatisticsChart.tsx to store the chart and have it dynamically loaded into the /Statistics page
 - changed ./ev-infrastructure/page.tsx to load the MapContainer, TileLayer, and StationMarker dynamically
 - changed ./usecases/[id]/page.tsx so that the Notebooks render dynamically

Overall First Load Reduction:
./privacypolicy: -128 kB
./statistics: -65 kB
./ev-infrastructure: -43 kB
./usecases/[id]: -95 kB

PR #2144, opened 2026-09-08
WD: Adding BundleAnalyzer Output to .gitignore

joshNewn01:
- Added the nodejs.html file that stores the output of BundleAnalyzer calls to .gitignore to not pollute future PRs


bettycheng14 (changes requested):
Hey @joshNewn01, good idea keeping generated reports out of the diff! Just two changes needed:

**1. Path won't match**

`next_webapp/.gitignore` patterns are already relative to `next_webapp/` (e.g. `/logs`, `/.next/`, `/out/`).So `/next_webapp/analyze/` would actually target `next_webapp/next_webapp/analyze/`.

→ Change it to `/analyze/`.

**2. `nodejs.html` is already tracked**

`next_webapp/analyze/nodejs.html` (~550 KB) is already committed on `master` from #2134. Adding it to `.gitignore` won't remove an already-tracked file.

→ run `git rm --cached next_webapp/analyze/nodejs.html` and commit the change.

Both changes are needed for the generated report to actually stay out of the repo.

next_webapp/.gitignore
```diff
@@ -20,7 +20,7 @@ yarn-error.log*
 # next.js
 /.next/
 /out/
-/next_webapp/analyze/
+/next_webapp/analyze/.
```
bettycheng14: should be `/analyze/`

bettycheng14 (approved):
Both issues from the earlier review are fixed. The `.gitignore` path now correctly matches, and the already-tracked file has been removed. Approving.

PR #2145, opened 2026-09-08
DS: Usecase Update UC00175

suprieyagtm10:
## Summary

Updated UC00175 in response to the Quality manager audit. 

## Changes completed

- [x] Updated notebook structure to match the current MOP use case template
- [x] Added a linked Contents section
- [x] Added Date Created and expanded prerequisite skills
- [x] Added a conclusion matching the template
- [x] Updated dataset handling to use reproducible project-relative paths
- [x] Removed redundant imports
- [x] Resolved warning and output issues
- [x] Added comments to code blocks
- [x] Standardised Australian English
- [x] Updated styling and layout to match the current template
- [x] Removed unnecessary HTML and JSON files


molliefernandez-mentor (changes requested):
Hi @suprieyagtm10 

**Author**: {Author/s}
**Level**: {Intermediate}
- Please change the datasets list so that it is Air quality - "about the dataset" with the dash in between 
- Also please update the dataset names to match their actual names
- The importing datasets section gives an error:
Image

- The data importing section needs to be pulling the data straight from the Melbourne open data site, not from zip files in the dependency folder. 
- The only dataset which should be in the DEPENDENCY folder is the "Bicycle Volume and Speed" (incorrectly labelled Bicycle count in the DEPENDENCY folder) - I actually think that you can call the data using its URL listed on the data site, and maybe nothing needs to eb in the DEPENDENCY folder. 
- All of the other datasets can be pulled using an API you do not need to have them in the DEPENDENCY folder. 
- The data cleaning process also looks super manual and like you are hard coding sections, I'm sure you can find a better way to do this. 
- Please use the template listed colours for the visualisations. 
- In the conclusion (above the sub headings) can you please talk about how the analysis has addressed the scenarios at the top
- Please update the use case index.

molliefernandez-mentor (changes requested):
Thanks @suprieyagtm10 

- Make sure you are deleting every file in the old READY TO PUBLISH folder for this use case including HTML and JSON files. 
- I am getting this error when I try and run the code

Image

molliefernandez-mentor (changes requested):
Thanks @suprieyagtm10 

- in the dataset section, just link the names of the datasets, don't add extra text. and then have the description after the dash like this:
[Bicycle Volume and Speed ](https://discover.data.vic.gov.au/dataset/bicycle-volume-and-speed)- This dataset contains cycling-flow records from bicycle counters across Victoria, including dates, times and speeds.

- Still getting errors running the code

Image

suprieyagtm10:
hi @molliefernandez-mentor , i have fixed the error and it should be fine now

molliefernandez-mentor (approved):
Thanks @suprieyagtm10 all looks good now! Just wait for @ashleyyy01's review then you can merge.

ashleyyy01 (changes requested):
Great work @suprieyagtm10!! Everything looks great, you are just missing one small thing.
- As per the template, Contents entries should be <a href="#section-1">Importing Packages</a>, but you are using the markdown instead: [Importing Packages](#section-1). Please update all entries in the Contents list to match the <a href="#id"> format, so that they are clickable and jump to the sections.

Another suggestion for section 4 which has 6 charts in it, maybe consider subheadings for each one like you have done for section 5 consisting of 5.1/5.2

Thank you looking forward to the final notebook.

suprieyagtm10:
Thanks for the feedback @ashleyyy01 .  
The final notebook is ready to be merged

PR #2146, opened 2026-09-09
WD: fix admin usecase category

bettycheng14:
### Issue found

Two admin use-cases pages still assumed the old flat `category_id` field on a use case. Category is an embedded ref `category: {id, legacy_id, category_name}` so both silently broke:

1. Edit page: category dropdown reset to "Select a category" on every open, even when one was already set — saving from that state would silently wipe it.
2. List page: Category column showed "—" for every row.

## Changes

1. `edit/[id]/page.tsx`: match the use case's category against the /api/categories dropdown options by either `legacy_id` (from Supabase-backed shape) or `id` (Mongo).
2. `use-cases/page.tsx`: read `category_name` directly off the embedded ref

### DIFF
--- a/next_webapp/src/app/[locale]/admin/use-cases/edit/[id]/page.tsx
+++ b/next_webapp/src/app/[locale]/admin/use-cases/edit/[id]/page.tsx
@@ -65,7 +65,6 @@ export default function EditUseCasePage() {
 
         setTitle(uc.title || "");
         setDescription(uc.description || "");
-        setCategoryId(uc.category_id ? String(uc.category_id) : "");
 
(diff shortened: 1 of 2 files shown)


hamizhassan16 (approved):
Traced this through end-to-end — the dropdown match logic is race-free (uc and categories are fetched together via Promise.all), the list page's category_name read matches what the API actually returns, and I checked that the save path (still sending category_id) is fine since the backend already resolves that into the same embedded category shape. Clean, correctly scoped fix. Approving.

HarshanaDeakinUni (approved):
LGTM

PR #2147, opened 2026-09-09
WD: Add Lighthouse CI Integration for Performance Monitoring

himeshsam93:
Implemented Lighthouse CI integration to automatically run performance audits on pull requests targeting the `staging` branch. The workflow builds the production frontend, audits key pages including Home, About, and Privacy Policy, checks performance budgets, and generates downloadable Lighthouse reports. This helps identify and prevent future performance regressions before changes are merged.


hamizhassan16 (changes requested):
Nice setup overall — the env-injection approach avoiding real secrets, the self-test harness for the assertion config, and being upfront about the current failing budgets (Home 63/100, Privacy failing LCP) are all good practice. One gap though: the docs and README describe .github/workflows/lighthouse.yml as already running on PRs against staging, but that workflow file isn't part of this PR (and .github/workflows/ doesn't exist in the repo at all yet). Could you add the actual workflow YAML so this is wired up, or update the docs to reflect that it's a follow-up step? Right now this PR only adds the local/config side.

himeshsam93 (commented):
I have addressed to the comment that i got for this PR. let me know if any further changes needed. Thank you!

hamizhassan16 (approved):
Confirmed .github/workflows/lighthouse.yml is now included and matches the documentation exactly — trigger conditions, no path filter, concurrency-cancel behavior, Node 22 + npm ci, the build/audit/upload steps, and the permissions model (read-only, no persisted credentials, safe for fork PRs since no secrets are required). Chrome resolution via CHROME_PATH is sound and degrades gracefully if the lookup ever comes back empty. Approving.

suba-t24 (changes requested):
Great setup! The mock variables and test scripts are well isolated. However, there are a few important issues we need to address before merging this:

1. Add pull_request: branches: [staging] to the workflow triggers so this actually runs on PRs.
2. In package.json, add --config=./lighthouserc.cjs to the lighthouse script to ensure it's loaded properly.
3. Change the test URL in lighthouserc.cjs to http://localhost:3000/en to avoid a redirect penalty on the performance score.
4. Verify that running with the dummy DB variables in lighthouse-env.cjs doesn't crash the homepage with a 500 error if Next.js tries to fetch data during SSR.

hamizhassan16:
Thanks for looking @suba-t24 . I went through these against the committed files (HEAD 6107f81af) and I don't think most of them apply:

1. The workflow already has this trigger - .github/workflows/lighthouse.yml starts with:
   on:
     pull_request:
       branches: [staging]
     workflow_dispatch:
   It was added in the same commit that addressed the earlier review. Might be worth a fresh look in case you were viewing an older version.

2. `lhci autorun` auto-discovers lighthouserc.cjs from the working directory (next_webapp/), so no --config flag is needed. The local-verification results in docs/lighthouse-ci.md show the budgets being enforced, which confirms the config is loading. Happy to add the explicit flag as belt-and-suspenders if you'd prefer, but it's not currently broken.

3. The config is consistently on port 3001 - startServerCommand starts the server on 3001 and every audited URL is http://localhost:3001/... They match, so there's no redirect. Switching the URLs to :3000 would actually break it since nothing listens on 3000. The URLs also already include /en, so no /-to-/en redirect penalty.

4. Reasonable to check. The docs note DB-backed home sections may render empty/error states by design, and the local run shows the Home page returning a real score (63/100, LCP 7,860ms) rather than a 500. The config also asserts http-status-code: error, so a 500 on any audited page would fail the workflow - it's already guarded.

Let me know if you're seeing something different in a specific file/line.

suba-t24 (approved):
@hamizhassan16 Good catch, you are completely right! I was looking at an outdated diff locally and missed the latest commit that already addressed all of these points. 
Everything looks perfectly wired up—the `pull_request` trigger is there, the config auto-discovery works flawlessly, and the ports/URLs match up nicely on `:3001` with the `/en` paths. I also see that the home page degrades gracefully without throwing 500s. 
Great work on this setup. Approving!

hamizhassan16:
> @hamizhassan16 Good catch, you are completely right! I was looking at an outdated diff locally and missed the latest commit that already addressed all of these points. Everything looks perfectly wired up—the `pull_request` trigger is there, the config auto-discovery works flawlessly, and the ports/URLs match up nicely on `:3001` with the `/en` paths. I also see that the home page degrades gracefully without throwing 500s. Great work on this setup. Approving!

Happy to help

PR #2148, opened 2026-09-09
WD: add GCP Cloud Run deployment pipeline via Cloud Build

HarshanaDeakinUni:
## Summary
Adds an end-to-end CI/CD pipeline that builds `next_webapp`, pushes the image to
Artifact Registry, and deploys to Cloud Run in `australia-southeast1`.

## Changes
- `next_webapp/cloudbuild.yaml`  ; 4-step pipeline: build → push to Artifact
  Registry → deploy to Cloud Run (`next-webapp-staging`) → apply
  domain-restricted access
- `next_webapp/Dockerfile` ; multi-stage build using Next.js `output: "standalone"`
- `next.config.mjs` ;  enable standalone output
- `supabase-Client.ts`lazy `getSupabase()` init to prevent build-time crashes(to be removed)
- `app/api/contact/route.ts` ;lazy `getTransporter()` init, same reason

## Secrets / config
Secrets are read from Secret Manager (regional replication,
`australia-southeast1`): `mongodb-uri`, `JWT_SECRET`, `SMTP_PASSWORD`,
`supabase-api-key`. No credentials are committed.

## Testing
- Docker image builds and runs locally
- Full Cloud Build trigger (`test0001`) passes end to end; service deploys

## Known limitations
- Public unauthenticated access is blocked by the org policy
  `iam.allowedPolicyMemberDomains`. Access is currently restricted to
  `domain:deakin.edu.au` as an interim measure; a load balancer is the likely
  long-term fix (pending cost approval).
- `NEXT_PUBLIC_APP_URL` is not yet wired in, needs the confirmed Cloud Run URL.
 
cheers


hamizhassan16 (changes requested):
The Docker/standalone-output migration itself looks right, and the lazy SMTP-transporter init is a good fix for Cloud Run's env timing (though the old commented-out code should just be deleted rather than left in). Two build-breaking issues though:

supabaseClient.ts drops the export const supabase in favor of getSupabase(), but 13+ files across the app still import { supabase } from '@/library/supabaseClient' (blogs, categories, gallery, home/*, upload, uploadImageToStorage, plus a few tests) — this will fail npm run build with a missing-export error. Either update all those call sites to getSupabase(), or keep a backward-compatible export.
activity-history/route.ts calls getSupabase() but never imports it, and the resulting variable isn't used anywhere in the file — looks like leftover debug code.

Can you run npm run build locally and confirm it's clean before this gets merged? Also curious if the GCP project id / Supabase URL / bucket name in cloudbuild.yaml are meant to be public in the repo.

HarshanaDeakinUni (commented):
Thanks for the review fixed them

activity-history/route.ts - removed, that was leftover debug code. Also deleted the commented-out transporter block and the old eager-client block in `supabaseClient.ts`.

Missing `supabase` export - as i have bypassed the build time resource check and we are decomissioning supabase i dident put much weight to this. fixed

Fixed with a backward-compatible export, - a Proxy that defers `getSupabase()` to first property access, so it doesn't reintroduce the module-load init crash the refactor was fixing. A plain `export const supabase = getSupabase()` would have put us straight back where we started.

On scope: blogs, categories, gallery and `profile/upload-image` are mine and move to MongoDB/GCS in my next PR i will push it thhrough once i resolve merge conflicts. 

Verified  `npm run build` is clean with no import warnings, `docker build` passes, and the container boots with no `.env` present, which is the condition that broke Cloud Build originally. also the service is building in the GCP so the build is successfull. test001 in the checks

On the identifiers in `cloudbuild.yaml` - all three are non-sensitive:

- GCP project ID is an identifier, not a credential; access is governed by IAM
- Supabase project URL is an endpoint, not a secret. These are `SUPABASE_URL` / `SUPABASE_API_KEY` with no `NEXT_PUBLIC_` prefix, so they're server-only and never reach the browser
- GCS bucket name is already behind Public Access Prevention and only reachable via our proxy API

The real secrets - `mongodb-uri`, `JWT_SECRET`, `SMTP_PASSWORD`, `supabase-api-key` - are in Secret Manager with regional replication in `australia-southeast1`. Nothing credential-bearing is committed.

Happy to fix anything else that i might have missed. Will push the next one soon.

hamizhassan16 (changes requested):
Good fix on the supabaseClient.ts Proxy — confirmed it keeps all 13 existing import { supabase } call sites (blogs, categories, gallery, home/*, upload, uploadImageToStorage, profile/upload-image) working while deferring client creation past import time, which was the actual Cloud Run issue. Requesting changes on two things before merge:

contact/route.ts — please delete the old commented-out transporter/env-check block instead of leaving it in. Dead code like this tends to confuse the next person editing the file (looks like two competing implementations).
cloudbuild.yaml — the GCP project id, Supabase project URL, and SMTP_USER address are hardcoded in plaintext in the repo. Can you confirm this is intentional for a staging deploy config, or should these move to substitution variables / a .env-style config that isn't committed? Even though the actual secrets go through Secret Manager, exposing the project id and infra layout in source isn't great practice.

lakindubd:
@hamizhassan16 as harshana mentioned in the meeting we had to make this workwround since the GCP team gave the account limited perms. It is currently being briefed to the juniours who will then have the opportunity to work on it next semester with the proper perms.

hamizhassan16 (approved):
Approving, changes will be made in future

lakindubd (approved):
Image

lgtm! confirmed builds passing,

PR #2150, opened 2026-09-10
WD: Task 6 - React Component Memoization

MilanKavindu2002:
### What I did
- Used React DevTools Profiler to check component re-rendering.
- Identified unnecessary HeroSlider re-renders during Dashboard updates.
- Added React.memo to HeroSlider.
- Added useCallback for the HeroSlider navigation functions.
- Added useMemo for the visible testimonials calculation.
- Profiled the LoginForm to check interactive form rendering.
- Tested the changes using React DevTools after optimization.
- Production build completed successfully.

### Testing
- Production build: Passed
- Tests: 186/187 passed
- 1 existing Contributors API test is failing and is not related to these changes.


champa1n73 (commented):
LGTM overall, the memoization and useCallback/useMemo changes should help reduce unnecessary re-renders, and the PR is small and focused. Before merging, please double-check that all useCallback/useMemo dependency arrays are complete, as missing dependencies could cause stale values or incorrect navigation behaviour. Also, confirm that HeroSlider props remain stable, or consider using a custom React.memo comparator when passing objects or arrays to avoid updates being skipped. If HeroSlider exposes a ref, make sure forwardRef and any imperative APIs are preserved so existing functionality doesn't break.

MilanKavindu2002:
I double-checked the useCallback and useMemo dependencies and they are correct. I also tested the HeroSlider after the memoization changes, including the slide navigation and swipe functionality, and everything is working correctly. HeroSlider does not use forwardRef or any imperative API, so I think no changes were needed for that.

hamizhassan16 (changes requested):
The memoization approach here is solid - switching handleNext/handlePrev to the setCurrentSlide((prev) => ...) functional form instead of reading currentSlide from closure correctly avoids a stale-closure bug that useCallback([]) would otherwise introduce, and memo(HeroSlider) is only effective because you paired it with stabilizing its onNext/onPrev props. Good understanding of the pitfalls here.

One thing to clean up before merge: the diff introduces inconsistent indentation in two spots - Dashboard.tsx's goToSlide (useCallback) line has an extra tab versus the surrounding code, and TestimonialsSection.tsx's useMemo body isn't indented relative to its opening line. Nothing functional, just asking for a quick formatting pass.

champa1n73 (approved):
The memoization work looks solid overall and the functional state updates in Dashboard.tsx avoid the stale-closure issue that would have caused incorrect navigation behavior; the HeroSlider memoization and the useMemo for visible testimonials are appropriate and do not show a functional bug. I don’t see any significant logic or security issues here, so this is in good shape to merge.

hamizhassan16 (approved):
The memoization approach here is solid - switching handleNext/handlePrev to the setCurrentSlide((prev) => ...) functional form instead of reading currentSlide from closure correctly avoids a stale-closure bug that useCallback([]) would otherwise introduce, and memo(HeroSlider) is only effective because you paired it with stabilizing its onNext/onPrev props. Good understanding of the pitfalls here.

Thanks for the formatting pass too - TestimonialsSection.tsx's useMemo is properly indented now. Approving.

PR #2151, opened 2026-09-10
WD: Migrate Blog, Gallery, Category & profile image uploads to MongoDB

HarshanaDeakinUni:
## Summary
Migrates Blog, Gallery, and Category API routes, plus the profile image upload path, from Supabase to MongoDB via the existing Mongoose models. Part of Phase 2 of the Supabase → MongoDB migration.

## Changes
- Blog routes (`src/app/api/blogs/`) rewritten against the `Blog` model via `dbConnect()`
- Gallery routes (`src/app/api/gallery/`) rewritten against the `GalleryImage` model
- Category routes (`src/app/api/categories/`) rewritten against the `Category` model
- Profile image upload path switched to the GCS proxy (`profile_img` on `User.profile`)
- Supabase Storage upload calls in this scope replaced with GCS (`mop-images` bucket)

## Conventions followed
- `created_by` stored as ObjectId, null fallback pending Auth migration
- `.populate()` for author/user lookups
- `$regex` for search/filter queries
- Explicit 404 on DELETE for missing records
-
## Notes
- The images were tested on the personal GCP bucket because deakin IAM policy won't allow external access.


champa1n73 (commented):
Nice work migrating the blog, gallery, and category routes and switching profile uploads to GCS. The changes are clear and mostly well-structured, but there are a few important issues to address before merging. The GCS upload flow should include proper validation and access controls, such as checking file type and size, sanitising filenames, and ensuring the correct bucket permissions or signed URLs are used. It would also be good to handle partial failures between GCS uploads and database saves to avoid orphaned files. The user-provided `$regex` search could also create performance risks, so consider escaping the input or using text indexes with appropriate limits. Finally, `created_by` should be properly validated and authenticated rather than falling back to `null`, and the `dbConnect` connection handling should be checked for reuse and errors. Adding tests for upload failures, database errors, and permission checks would also help improve reliability and security.

hamizhassan16 (changes requested):
The migration itself is well done - regex-escaped search filters, real WebP re-encoding via sharp so the .webp extension is accurate, deleteImageFromGCS correctly skipping legacy Supabase URLs, and DELETE routes cleaning up storage before removing records.

But there's a real bug that needs fixing before merge: in both blogs/[id]/route.ts and gallery/[id]/route.ts, PUT uploads a new cover image and overwrites the stored URL, but never deletes the OLD image from GCS first. Every edit that replaces an image leaks the previous one into the bucket permanently (it's gone from the DB the moment it's overwritten, so there's no way to clean it up after the fact). DELETE handles this correctly in both files - PUT needs the same "capture old path, upload new, then best-effort delete old" pattern.

Two smaller things:
1. .gitignore now has a line that's just ".gitignore" - the file ignoring itself, looks accidental, please remove.
2. /api/images/[...path]/route.ts is public/unauthenticated and will serve any object in the bucket, not just ones under blogs/covers/, gallery/, profiles/. Worth adding a prefix allowlist so the proxy can't be used to read arbitrary bucket objects if anything else ever gets stored there.

Also noticed an unrelated notebook rename (...Elderly_Social_Support_Networks_Sprint1.ipynb) with zero content changes riding along - is that intentional for this PR?

Requesting changes for the orphaned-image bug primarily.

hamizhassan16 (approved):
All three points are fixed, and fixed well:

1. The orphaned-image bug is resolved - both blogs/[id]/route.ts and gallery/[id]/route.ts now capture the previous cover/image URL before overwriting, and only delete it from GCS after the new one is successfully saved to the DB. That ordering is actually better than what I'd suggested - it means a failed save never leaves the record pointing at a deleted file.

2. The self-referential .gitignore line is gone.

3. /api/images/[...path] now has both a ".." traversal guard and an explicit ALLOWED_PREFIXES allowlist (blogs/, gallery/, categories/, profiles/), so the public proxy can't be used to read arbitrary objects from the bucket.

Approving.

hamizhassan16 (changes requested):
Catching one more thing before this merges: there's a stray root-level package-lock.json in this PR (not next_webapp/package-lock.json - an actual new file at the repo root). Checked its content:

{
  "name": "MOP-Code",
  "lockfileVersion": 3,
  "requires": true,
  "packages": {}
}

This doesn't exist on master and doesn't do anything useful - it looks like `npm install` got run from the repo root instead of `next_webapp/` at some point, which auto-generates this empty stub lockfile. All package management for this project lives under next_webapp/, so this shouldn't be tracked at repo root.

Could you remove it (`git rm package-lock.json` from the repo root) before this merges? Requesting changes for that - everything else in this PR is solid.

HarshanaDeakinUni (commented):
Thanks you both, everything raised is now addressed.

**From @hamizhassan16**

- **Root `package-lock.json`** - removed , good catch on that!!
- **Orphaned image on PUT** - fixed in both `blogs/[id]` and `gallery/[id]`: capture the previous URL, upload, save, then delete. 
- **Self-referential `.gitignore` line** - removed.
- **un necesary playground file** - removed, was leftover work from last trem.
- **Public image proxy** - `/api/images/[...path]` now has a `..` traversal guard plus an `ALLOWED_PREFIXES` allowlist (`blogs/`, `gallery/`, `categories/`, `profiles/`).

**From @champa1n73**

- **Tests** - added unit tests for the scope: 87 tests across 5 suites covering blogs, gallery and upload. They cover upload failures, database errors and permission checks, plus regression cover for the orphaned-image bug, including that the old object is deleted only after `save()` succeeds. GCS and Mongoose are mocked, so nothing touches a real bucket or database.
- **Partial failure between upload and DB save** - POST on blogs and gallery now deletes the uploaded object if the insert throws, so a failed save can't orphan a file.
- **File type and size validation** - already in place: every upload route checks an allowed MIME list and a 5 MB limit before anything reaches GCS.
- **Filename sanitising** - filenames are generated server-side (`gallery-<userId>-<timestamp>.webp`), so no user input reaches them. The one caller-controlled segment, `folder` in `/api/upload`, is restricted to `[a-zA-Z0-9_-]` and rejects path traversal.
- **`$regex` input** - already escaped in every search path.
- **Bucket permissions / signed URLs** - the bucket is private and images are served through the authenticated server-side proxy, so no public access or signed URLs are needed.
- **`dbConnect`** -  already caches the connection and clears the pending promise on failure so the next request retries.
image

hamizhassan16 (approved):
Confirmed the root package-lock.json is gone - diff against master is now clean for that path. Also noticed you added real unit test coverage for the blog/gallery/upload routes since I last looked, including tests that specifically target the save-then-delete ordering for image replacement (deletes the old object only after save() succeeds / does not delete anything when save() fails) - good, targeted regression coverage for exactly the bug that mattered.

Approving.

PR #2152, opened 2026-09-11
DS: Usecase Update UC00009

AAYUSH2201:
Reworked UC00009 moved it from READY TO PUBLISH into FINALISED. The notebook was reformatted, its ten flat sections regrouped into six with subsections so every code cell has an explanation before it and an interpretation after it, and the code cleaned — a misspelled column, a paging loop that could not terminate on a failed request, duplicate imports and two unhelpful error paths are all fixed. On presentation, `geopy` replaces the hand-written haversine, the maps are framed with `fit_bounds`, markers are coloured by category, and it now runs top to bottom without prompting for input. Renamed to "Melbourne Child Friendly Places Analysis 2023" and the old `.html` and `.json` were deleted with it.


VedS8 (changes requested):
Hey Aayush, really solid rework! Personally, I could only find two things:

1. The child-friendly map in cell 55 colour-codes 11 categories, but the only legend is a printed text list above the map rather than an actual legend on the map itself. Could you please add an on-map legend so the categories are easy to understand without having to scroll up?

2. Could you please also update the Use Case Index to include the changes from this use case?

Other than that, everything looks great and runs clean. Great work!

molliefernandez-mentor (changes requested):
Hi @AAYUSH2201 

- Change the scenario to this:
As a parent new to the city, I want to see which parks, schools, pools and other child friendly facilities sit near a given address, and how long it takes to reach the one I choose, so that I can decide where to go without knowing the city already. 

- Then add the below: to the start of the background section
The City of Melbourne publishes every landmark and place of interest within one of its open datasets. It covers the whole city and all types of sites, from car parks to sports facilities to places of worship. This use case will work to narrow down the data and filter it to the categories that are important to families with children. 

- All the folium maps look like this:
Image

- please make sure that the icons in your folium maps use the colours listed in the template (not the green through, it doesn't stand out well against the green of the maps)
- in the conclusion can you please add a section above the sub sections where you talk about how the use case has addressed the problem statement in the scenario.

AAYUSH2201:
. The child friendly map now has a legend drawn onto the map itself using the same `add_legend` helper as UC00026, and the markers have moved off green onto the template colours. I've also updated the Scenario and Background to your wording, added a section at the top of the Conclusion and moved UC00009 into the Finalised table in the use case index. The maps were also coming back as blank "access blocked" tiles from OpenStreetMap, so the basemap has changed and the notebook has been updated according to the feedback. Re-requesting review after the appied changes. @molliefernandez-mentor @VedS8

molliefernandez-mentor (changes requested):
Hi @AAYUSH2201, the first 3 folium maps work great! but the last one looks like this :(

Image

**Author**: {Author/s}
**Level**: {Intermediate}

AAYUSH2201:
I've zoom across all four maps, and I've also updated the use case details to the template format.  Re-requesting review after the appied changes. @molliefernandez-mentor

VedS8 (changes requested):
Hey Aayush, great work! I think there's just one thing left that I found.

The map colours (`MOP_PURPLE = '#D252B9'` and the `MARKER_SWATCHES` dict in cell 18) don't match the template hex codes. Could you please update them to the actual template colours, except for green, since Mollie mentioned that green markers are difficult to see against the map?

Everything else looks great and runs clean. Thanks!

molliefernandez-mentor (commented):
Looks great @AAYUSH2201 

Can you please just put brackets around your name: "Updated: Trimester 2, 2026 (Aayush Patel)"

Agreed with Ved to please use the template colours, once you use them all you can then move onto other colours. And you can use the green because the map you have is fairly grey - just change the legend text and border to black if you use the green.

AAYUSH2201:
Updated the name in header details using brackets and updated with black and green color in folium maps.  Re-requesting review after the appied changes. @molliefernandez-mentor @VedS8

VedS8 (changes requested):
Hey Aayush, close, just one thing left.

The colour palette in cell 18 still isn't using the template hex codes. Could you please replace the five custom colours with the correct template colours?

- Purple: #9333EA 
- Orange: #F97316
- Blue: #2563EB
- Yellow: #EAB308
- Pink: #EC4899

Once all six template colours are used, the remaining categories can use additional distinct colours as needed.

Everything else runs clean with no errors. Thank you!

molliefernandez-mentor (approved):
thanks @AAYUSH2201 all good from me, just make the final change @VedS8 has suggested and get his approval and you can merge!

AAYUSH2201:
The six template colours now updated. Re-requesting review after the appied changes. @VedS8

VedS8 (approved):
Great work Aayush, happy to merge!

PR #2153, opened 2026-09-13
WD: add admin workflow test suites

bettycheng14:
This PR adds Cypress E2E coverage for the admin CRUD workflows. All API callsare stubbed via `cy.intercept()` and auth is faked via `localStorage` without backend.

### New test files

**1. `cypress/e2e/admin_workflow.cy.ts`**
- Access control — no user / tokenless / non-admin / admin guard redirects
- Categories — list, add, edit, delete (cancel + confirm)
- Contributors — add (happy path + 3× PR #2138 regressions), edit (pre-fill + `project_lead` fallback), delete
- Blogs — list, add (empty validation + CKEditor API submit), edit

**2. `cypress/e2e/usecases.cy.ts`**
- Add — upload zone state, rejects wrong type / malformed JSON, valid upload, remove, full submit (verifies `content` is notebook JSON)
- Edit — tri-state: untouched (omits `content`), removed (`content: null`), replaced; malformed replacement rejected without losing existing state
- Delete — `category_name` sourced from embedded ref (regression for `toUseCaseDTO` shape), cancel + confirm

### Fixtures added
| File | Purpose |
|---|---|
| `fixtures/sample.ipynb` | Valid notebook for upload tests |
| `fixtures/malformed.ipynb` | Invalid JSON with `.ipynb` extension |
| `fixtures/not-a-notebook.txt` | Wrong file type rejection |
| `fixtures/cover.png` | 1×1 px PNG for blog cover upload |

### Verification
- ✅  All Cypress E2E tests pass (`npx cypress run`)
- ✅ `npm run build` completes successfully


hamizhassan16 (approved):
Checked the two pre-existing test files this PR modifies since those are the risk area in an otherwise purely-additive PR. Both changes are legitimate fixes, not regressions - verified the auth.cy.ts error-message updates against the actual string in src/lib/apiFetch.ts, and confirmed navigation.cy.ts's header-scoped selector is needed because Header.tsx and Footer.tsx both render an element with the same aria-label, which would have made the old unscoped selector ambiguous.

No hardcoded secrets, no .only/.skip left in, fixtures look fine. Approving.

PR #2154, opened 2026-09-14
DS: Usecase Update UC00110

Sohamdeakin:
Changes done to the UC00110 

- I updated the notebook structure to better match the current MOP template.
- I improved the Scenario, Background, Datasets, Contents and Conclusion sections.
- I removed the old Google Colab/API key setup and cleaned unnecessary imports.
- I updated the data loading so the datasets can be downloaded directly from the City of Melbourne public API.
- I fixed the pedestrian dataset field change from timestamp to sensing_date.
- I corrected venue category and filtering issues.
- I added R² and MAE metrics to the Linear Regression section.
- I improved the chart colours and overall presentation to match the MOP template.


VedS8 (changes requested):
Hey Soham, good start on this one! There’s a fair bit to fix, but most of it should be pretty straightforward.

1. **Title:** The title (cell 1) isn't wrapped in the `usecase-header-bar` div. It's just using `usecase-title` with no green bar behind it. Could you please wrap it like the other approved use cases?

2. **Date Created:** The details block (cell 2) is missing the "Date Created" field. Could you please add it in with the original creation trimester?

3. **Bold text:** "What this use case will teach you" and "Background" (cells 5 and 7) are missing the `<b>` tag that Scenario and Contents have. Could you please add this so they all render consistently?

4. **Contents heading:** The Contents heading (cell 10) is using `class="usecase-contents"` instead of `usecase-unnumbered-heading`, and it doesn't have an id. Could you please update the class and add an id like the other headings?

5. **Dataset bullets:** In the Datasets section (cell 9), all six bullets are written as `-[Name](link)` without a space after the dash. Could you please add the space after `-` so Markdown renders them properly as a list?

6. **Unused datasets:** "City Activities and Planned Works" and "COM population forecasts by small area for 2020-2040" are listed as datasets but don't appear to be downloaded or used anywhere in the code. Could you please either wire them into the analysis or remove them from the list?

7. **Colour palette:** The colours defined in cell 17 (`color_d`, `color_l`, `color_median`, etc.) don't use the template hex codes and are currently a custom palette. Could you please update the plots to use the template palette?

8. **Venue Capacity heading:** Cell 39 is currently just plain text, "Venue Capacity Over Time", with no div or heading class, so it won't get the proper styling. Could you please wrap it as a heading? It looks like it should be a 7.1 subheading under Venue Trends.

9. **Conclusion heading classes:** In the Conclusion (cell 56), the three subheadings are using `usecase-subsection-heading`, which is missing the hyphen. Could you please change this to `usecase-sub-section-heading` so the template styling applies?

10. **Conclusion wording:** Those same three subheadings should match the standard template wording: "What we achieved in this analysis", "What we learned from this analysis" and "Observations for further opportunities". Could you please update them from the current shortened versions?

11. **References:** The References section (heading 12, cells 57 and 58) should be removed, same as the other use cases in the repo. Could you please remove this section as well?

12. **HTML/JSON files:** Could you please remove any HTML or JSON files that were generated for this use case? These shouldn't be included in the final PR.

13. **Index:** Could you please update the use case index file to include this use case and make sure the entry matches the correct title and details?

The notebook does run clean with no errors, and the Scenario, Background and What You'll Learn setup at the top reads well. Just these bits to tidy up. Thank you!

molliefernandez-mentor (changes requested):
Hi @Sohamdeakin 

- Move the use case into the FINALISED folder
- update the use case index
- update the CSS to the version in the template
- apply the header bar to the use case title
**Author**: {Author/s}
**Updated**: {Trimester #, YYYY} {Student Name}
**Level**: {Intermediate}
- put the use case detail in the same box as the title
- include a date created
- remove the updated by section
- Make sure you are using only the use case template listed colours. 
- Check the data sets list - some of the links don't work or don't link to the information section or have different names
- make sure they are dot pointed with dots not dashes
- the contents heading needs to be the unnumbered heading div tag
- Can you please change the headings int he contents to be inline with the other use cases like "set-up" should be "importing libraries" and "view datasets" should be like "retrieving the datasets" please update them all to be more like the other use cases. 
- There needs to be explanations after each heading as so what you are doing and why. not just 1 or 2 pointless dot points. 
- the conclusion should be an unnumbered heading
- remove the references section
- in the conclusion add a section above the sub headings about how the use case has addressed the problem statement outlined in the scenario. 

This use case still needs a lot of work..

Sohamdeakin:
Hi @molliefernandez-mentor & @VedS8 
I have updated UC00110 based on the latest feedback:

- updated the template, CSS and title/header section
- fixed the use case details and added the created/updated information
- cleaned the datasets section and removed the unused datasets
- updated the section headings and added explanations about what is being done and why
- changed the plots to use the template colours
- updated the conclusion and removed the references section
- moved the final notebook to the FINALISED folder
- removed the old UPDATE NEEDED copy and the generated HTML/JSON files
- updated the Use Case Index for UC00110

The notebook is running cleanly end to end. Please have another look when you get a chance, thank you!

molliefernandez-mentor (changes requested):
Hi @Sohamdeakin 

- Make sure you delete the use case from its original table in the use case index.
- Change updated to this "Updated: Trimester 3, 2026 (Soham Chodanekar)"
- section 7.1 is not listed in the contents list

Otherwise it all looks good!

Sohamdeakin:
Hi @molliefernandez-mentor & @VedS8 I have done the final UC00110 changes:

- removed the old index entry
- updated the trimester/details
- added section 7.1 to the Contents

VedS8 (changes requested):
Hi Soham, sorry about this one, it's on me. I told you to fix that missing hyphen in my previous feedback, but pointed you at the wrong class.

The three Conclusion subheadings (cell 56) should be `usecase-body` with a `<b>` tag, not `usecase-sub-section-heading`. Could you please switch those three back to `usecase-body`?

Also, just to confirm with Mollie, should the Updated field be T2 2026 instead of T3 2026? The other use cases use T2 2026.

Everything else is sorted and runs clean. Thank you!

molliefernandez-mentor (approved):
Thanks @Sohamdeakin, yes agree the updated date should be Trimester 2 2026.

Once you have made the changes suggested by @VedS8 you are welcome to merge!

Sohamdeakin:
Updated the final UC00110 feedback:

- changed the three Conclusion subheadings back to usecase-body with bold text
- updated the Updated field to Trimester 2, 2026 as confirmed by Mollie

Everything else remains unchanged. Please have another look when you get a chance, thank you!

molliefernandez-mentor:
Thanks Soham, wait for @VedS8's final review then you can merge

VedS8 (approved):
Great work, Soham. Good to merge!

PR #2155, opened 2026-09-14
DS: Usecase Update UC00104

AtishayDeakin:
Restructured the entire notebook (167 → 199 cells) to the current template. The analysis itself is preserved, not rebuilt, this was a reformat and cleanup pass, not a redesign.

Template alignment
- Header bar, metadata, Scenario / What this teaches / Background / Data Sets (7 sources linked) / Contents
- 14 numbered sections with subsections, description under every heading
- Removed References section; merged the old "Part 2" restart into continuous numbering

Content cleanup
- Fixed genuinely corrupted explanatory text (garbled/mashed words)
- Removed leftover HTML-comment draft explanations sitting next to their final versions
- Consolidated 3 near-identical routing code blocks down to 2 (intro + final version)
- Flagged rather than silently deleted 3 fetched-but-unused datasets (microclimate sensors, monthly pedestrian counts, CSV street names), noted in Conclusion as a future opportunity

Bug fixes
- Stray leading whitespace causing a `SyntaxError` in one cell
- Two local `.geojson` dependencies that don't exist in the repo — replaced with live API fetches
- `CRS` error the API fetch introduced (`from_features()` doesn't set CRS the way `read_file()` does) — fixed with explicit `set_crs(epsg=4326)`
- Missing `API_KEY_ORS` — documented as a required separate credential from `API_KEY_MOP`

Data-accuracy fix
- Conclusion referenced a stale regression result (R² 0.52) from a previous run. Since Section 3.4 always filters to "the last month" relative to execution time, this model's reported performance is inherently a snapshot, not a fixed benchmark, reworded the two "What we achieved/learned" summaries to point at the live Section 11.3 output instead of a hardcoded figure, so it can't go stale on the next re-run either.


AtishayDeakin (commented):
@molliefernandez-mentor and @VedS8,

I haven't added the entry to use case index as of now because I am waiting for the final approval and before that there will be conflicts so I am just waiting till the final approval.

VedS8 (changes requested):
Hey Atishay, good work on this one, just a few things to sort:

1. Cells 25, 29, 158 and 169 all call `.head` without the parentheses, so they print `<bound method NDFrame.head of ...>` and dump the whole dataframe instead of a clean preview. Could you add the `()` to all four?

2. Four spots aren't using the template colours. The footpath steepness map (cell 174) uses a six-colour blue-to-red gradient with no legend, the two pedestrian heatmaps (cells 85 and 88) use Folium's default heat gradient, and the correlation heatmap (cell 127) uses `cmap='coolwarm'`. Could you bring all of these in line with the template palette? The gradients should still be based on the primary green rather than the default library colours.

3. Title/filename: The title still says “...Melbourne's Urban Areas 2024”, but the filename has no year in it. Could you add “2024” to the filename so it matches the title?

Everything else is solid. Runs clean start to finish, CSS and heading structure are current, the conclusion ties back to the scenario properly, and the dataset list and API key requirement are documented. Nice work so far!

molliefernandez-mentor (changes requested):
Hi @AtishayDeakin 

This use case is too large and can't be opened in GitHub, the size needs to be reduced. 
Image

I get all sorts of errors when trying to open it:
Image

and errors when trying to run the code:
Image

**Author**: {Author/s}
**Level**: {Intermediate}

AtishayDeakin (commented):
Hi @molliefernandez-mentor and @VedS8,

I have made the requested changes.

VedS8 (changes requested):
Hey Atishay, a couple of things still outstanding:

1. **Use case details:** Could you please update them to follow this format?

    **Author:** {Author/s}
    **Created:** {Trimester #, YYYY}
    **Updated:** {Trimester #, YYYY} {(Student Name)}
    **Time to Complete:** {90} mins
    **Level:** {Intermediate}
    **Pre-requisite Skills:** {Python, and any other required skills}

2. **Notebook execution:** Cell 189 has a saved `ValueError: No API key was specified` for OpenRouteService, and every cell from 192 onwards hasn't been executed. Could you please set `API_KEY_ORS` and run the notebook from start to finish before saving it?

Good progress on the other feedback. Once these two are sorted, it should be ready. Thanks!

AtishayDeakin (commented):
Hi @VedS8,

I have done the final requested changes, hope it is good to go now.

molliefernandez-mentor (changes requested):
Hi @AtishayDeakin 

- Can you please update the use case index
- please make sure the visualisations are using the colours listed in the template. 
- Make sure all your maps have legends
- all the folium maps have this issue:

Image

- I am getting this error when running the code

Image
Image

PR #2163, opened 2026-09-17
WD: perf optimize MongoDB search pagination (Task 8)

Kangbunbun:
## Summary

Completes Lakindu’s Task 8 by moving `/api/search` pagination into MongoDB and adding the missing `category.legacy_id` index.

## Changes

* Replaced fetching all matching documents and slicing in JavaScript with `countDocuments()`, `skip()`, and `limit()`.
* Preserved existing filters, sorting, safe-field projection, response format, and out-of-range page handling.
* Added the `category.legacy_id` index to the UseCase schema.
* Fixed a sort-memory-limit error found during live testing. For MongoDB error 292, the query retries using an aggregation pipeline that projects safe search fields before sorting and pagination.
* Updated pagination tests and added sort-fallback regression tests.

## Validation

* 4 targeted Jest suites passed: **66/66 tests**.
* Live API checks: pages 1 and 2 returned 2 results each.
* After the fix, `page=999&pageSize=2` returned page 20 with 2 results, total 40, and `hasNext: false`.
* Read-only database inspection confirmed `category.legacy_id_1` exists on `mop_database.usecases`.
* Production build completed with exit code 0.
* `git diff --check` passed.

## Existing issues / limitations

* The existing package-lock.json contains invalid JSON.
* The build reports invalid ESLint options and skips TypeScript validation; neither is claimed as passing.
* No performance benchmark was conducted.
* Changes are limited to the search route, UseCase schema, and search tests.


hamizhassan16 (approved):
This is a solid, correctly-reasoned fix. Verified SEARCH_PROJECTION is a valid shape for both .select() and $project, and that every whitelisted sortBy value (created_at/updated_at/title) is present in that projection, so the aggregate fallback's $sort never operates on a field the $project stage stripped out.

The error-292 fallback is the right fix for the actual problem (QueryExceededMemoryLimitNoDiskUseAllowed on shared Atlas tiers) - projecting to safe fields before $sort is what actually reduces the in-memory sort footprint, which a plain find().select().sort() doesn't get you. The unknown-safe error narrowing before reading .code is careful, correct TypeScript.

Good regression coverage too - the sort-memory-limit test actually mocks a {code: 292} rejection and asserts the resulting aggregate pipeline shape (including that content/content_file_id stay excluded), plus a failure-cascade test for when the fallback itself fails. The new category.legacy_id index matches what the filter builder actually queries on.

One thing worth a quick note, not a blocker: the 292 path is retry-after-failure rather than the primary path, so any query that's going to hit it pays for a failed find() first. Given most searches on this collection are probably well under the sort memory limit, keeping find() as the fast common path and only paying aggregation overhead on the rare large-result-set queries seems like a reasonable trade-off - just flagging in case it was accidental rather than deliberate.

Approving.

Kangbunbun:
Thank you for the detailed review and approval.

I appreciate the note about the error-292 retry path. I kept the normal `find()` query as the primary path so typical searches can use the simpler and faster query flow, while the aggregation pipeline is only used when the shared Atlas sort-memory limit is actually reached.

I agree that this means affected queries pay for the initial failed attempt. If this becomes a frequent production issue, making the aggregation path primary for those cases would be worth revisiting.

Thanks again for verifying the projection, fallback behaviour, index, and regression coverage.

PR #2166, opened 2026-09-18
WD: Fix admin workflow Cypress tests after contributor section merge

bettycheng14:
Follow-up on #2138 for fixing several tests with documented gaps.

### 1. Stale assertions flipped 

- Updated 4 tests from not.exist → be.visible and renamed them accordingly (Project Lead, Project Team, and Data Science Assistant Team Lead)
- Edit project_lead test updated to expect "👑 Project Lead" label instead of the "Select type" placeholder

### 2. Selector fixes for dual mobile/desktop DOM

- Empty-state assertion scoped to cy.get('table') — avoids the hidden md:hidden mobile container
- Delete buttons use .last() instead of .first() — mobile card button renders first but is hidden; desktop table button is last
- Modal name assertion scoped inside .fixed.inset-0 — avoids matching contributor name in the hidden mobile card

### New test

- Added missing edit contributor submit test: verifies PUT request body, success toast, and redirect

### Verification
- All 28 tests passed with `npx cypress run --spec "cypress/e2e/admin_workflow.cy.ts"`


hamizhassan16 (approved):
Went through this against the actual current state of master rather than taking the comments at face value:

- The dropdown assertion flips (project_lead, Project Team, Data Science Assistant Team Lead) match what #2138 actually shipped in types/contributor.ts - confirmed all three are really there now.
- The delete-button .last() fix is correct - checked the component source and the mobile card (md:hidden) genuinely renders before the desktop table (hidden md:block) in DOM order, so .first() really was grabbing a hidden, unclickable button.
- The new submit test (PUT body, toast, redirect) is solid new coverage, not just patching over old assertions.

Approving.

PR #2167, opened 2026-09-19
WD: fix(images): add onError placeholder fallback across all image render sites

lakindubd:
What this does

The Supabase Storage bucket that hosted our legacy images is being shut down. Without remediation, ~38 images across the gallery, blogs, categories, and user profiles would render as broken icons the moment Supabase goes dark. This PR is the code-side half of that fix: it adds an onError placeholder fallback at every image render site so any dead image degrades to a neutral placeholder instead of a broken-image icon.

the data migration is already done, and is NOT part of this PR. I ran updateMany directly against production Atlas to repoint every Supabase Storage URL to the local placeholder /images/category-placeholder.png. This PR contains only the code changes; the DB is already remediated. Evidence of the data migration is in the description block below.

image


HarshanaDeakinUni (approved):
This covers the discussed scope and lgtm

PR #2168, opened 2026-09-20
WD: Complete backend migration off Supabase repoint blog reads to MongoDB, remove Supabase client

lakindubd:
What this does

This completes the backend's migration off Supabase. It repoints the last two remaining Supabase reads at MongoDB and removes the now-unused Supabase client and dependency. After this PR, **there is zero live Supabase code anywhere in `src/`**  the backend is fully on MongoDB/GCS.

Every API route was already migrated in earlier work; the only remaining Supabase usage was two blog read paths that live *outside* `src/app/api/` (in the sitemap and blog-detail page), which is why they were missed until now.

Updated the env template as well

Changes

**1. `src/app/sitemap.ts`  sitemap blog entries now read from MongoDB**
- `getBlogEntries()` was calling `supabase.from("blogs")`; it now reads from the Mongo `Blog` model via `dbConnect()`, mirroring how `getUseCaseEntries()` in the same file already works.
- Blog URLs in the sitemap now use the Mongo `_id` (ObjectId string), matching the real `/blog/[id]` routes previously they used the old numeric id.

**2. `src/app/[locale]/blog/[id]/page.tsx` blog-detail SEO metadata now reads from MongoDB**
- `getBlogForMetadata()` (used by `generateMetadata`) was calling `supabase.from("blogs").eq("id", Number(id))`.
- **This fixes a real bug:** blog ids are Mongo ObjectId strings now, and `Number(<objectid>)` is always `NaN`, so this metadata lookup was silently broken  every blog page was falling back to generic SEO metadata with no cover image. It now validates the ObjectId and does `Blog.findById(id)`, so OpenGraph title/description/image work again.

**3. `src/library/supabaseClient.ts` deleted + dependency removed**
- With the two reads above repointed, nothing live imports the Supabase client anymore, so it's deleted.
- Removed `@supabase/supabase-js` from `package.json` dependencies (lockfile regenerated).


HarshanaDeakinUni (approved):
Lgtm, approving

PR #2169, opened 2026-09-20
DS: Usecase Update UC00025

Kush-Patel32:
Addressed quality review feedback including template formatting, historical fill-level features, time-aware model evaluation, data cleaning, spelling and presentation improvements, final prediction visualisation, conclusion and references.


molliefernandez-mentor (changes requested):
Hi @Kush-Patel32 

- Make sure you delete the previous version of the use case in the READY TO PUBLISH folder. 
- delete this sentence "The analysis and code begin below."
- there are a few sub headings that are not in the contents list - like "1.3.1 Pearson Correlation Analysis"
- Make sure all of the visualisations are using the template colours - currently the section 3.3, 3.4, 3.5 ones are not. 
- delete the references section at the bottom
- make sure the conclusion have the 3 sub headings as listed in the template.

VedS8 (changes requested):
Hey Kush, good progress. Just a few things still need fixing:

**Use Case Index:**

1. The new UC00025 row is broken. It's `[File Link] https://...` with no parentheses around the URL, so that link won't render as a clickable link at all, it'll just show as plain text.
2. The link itself points to `blob/Kush-Patel32-UC00025/usecases/master/UC00025_Smart_Bin_Waste_Collection_AI_Application.ipynb`, that's your branch name and "master" mixed into the path in the wrong order. It should be `blob/master/usecases/FINALISED/UC00025_Smart_Bin_Waste_Collection_AI_Application.ipynb`, same pattern as every other row.
3. Could you please also remove the old T1 2024 entry for UC00025 from the Use Case Index? Since it's now in the FINALISED table, the old entry is no longer needed, and removing it will ensure the use case only appears once in its correct location. Thank you!

**Notebook:**

4. The three Conclusion subheadings (cell 84) use `usecase-sub-section-heading`, but it should be `usecase-body` with a `<b>` tag for these. Could you please switch them over?

Everything else is solid: contents now includes 1.3.1, the References section and "analysis and code begin below" sentence are both gone, and the colours in the correlation heatmap, model comparison chart and confusion matrix are all green-based now. Runs clean, no errors. Thank you!

molliefernandez-mentor (changes requested):
Thanks @Kush-Patel32 

- Please change it to Trimester 2 instead of T2
- delete this sentence: "This use case uses public datasets relating to monthly waste collection and historical smart-bin sensor readings."
- a lot of stuff happens under 2.1 which isn't well explained, can you ease add more explanation and or sub sub headings to explain what is happening. 
- delete the empty code box and the end of the use case

PR #2175, opened 2026-09-21
DS: MOP Data Dashboard: strengthen dataset coverage

s224251967-hub:
The dashboard omitted UC00023 when its Data Sets heading and links shared a cell, and UC00053 when dataset IDs were passed through variables. Separately, registry pushes rejected by protected master stopped Pages publication after a successful build.

This change expands extraction for those notebook formats, counts the complete FINALISED inventory separately from dataset coverage. Workflow code and notebook input are pinned to the same selected commit. Registry persistence failures no longer block publication.


liajohn30 (approved):
Reviewed and tested the dashboard locally. Confirmed that the FINALISED Use Cases count is correctly showing 20 and matches the current FINALISED inventory on master. Everything looks good from my side. Thank you!

